PR #2, opened 2020-07-20
a few fixes to copy-json-to-bipf.js

staltz:


copy-json-to-bipf.js
```diff
@@ -1,6 +1,6 @@
 var pull = require('pull-stream')
 var FlumeLog = require('flumelog-offset')
-var FlumeLogRaf = require('../')
+var FlumeLogAligned = require('flumelog-aligned-offset')
```
staltz, line 3: `require` of `../` didn't make sense to my local folder structure when cloning this repo

copy-json-to-bipf.js
```diff
@@ -11,15 +11,13 @@ var block = 64*1024
...
+var log2 = FlumeLogAligned (process.argv[3], {block: block})
 
-var log2 = FlumeLogRaf (process.argv[3], {block: block})
-var a = [], length = 0
 pull(
   log.stream({seqs:false, codec: json}),
   pull.map(function (data) {
     var len = binary.encodingLength(data)
     var b = Buffer.alloc(len)
-    length += b.length + 4
```
staltz, line 22: This variable `length` (and `a`) were written to, but never **read**, so I figured we could delete it

example.js
```diff
@@ -1,4 +1,3 @@
-const util = require('util')
```
staltz, line 1: Never used

copy-json-to-bipf.js
arj03: right, this was because it was copied over from flumelog-aligned-offset repo and never properly fixed.

example.js
arj03: Yep, it was there because I've used it pretty print some json values. Fine with this change.

arj03:
Perfect! Thank you :)

PR #6, opened 2020-11-05
New query interface

arj03:
This is WIP

Examples:

```javascript
  const mix = '@ye+QM09iPcDJD6YvQYjoQc7sLF/IFhmNbEqgdzQo3lQ=.ed25519'
  const mixy = '@G98XybiXD/amO9S/UyBKnWTWZnSKYS3YVB/5osSRHvY=.ed25519'
  
  var q = query(db, query.type('post'))
  q = q.AND(query.channel("solarpunk"))
  q = q.AND(query.OR([
    query.author(mix),
    query.author(mixy)
  ]))

  //console.log(JSON.stringify(q.op, null, 2))
  
  q.paginate(0, 10, (err, msgs, paginator) => {
    console.log(msgs.length)
    paginator.next((err, msgs) => {
      console.log(msgs.length)
      paginator.next((err, msgs) => {
        console.log(msgs.length)
        console.log(paginator)
      })
    })
  })
```

or

```javascript
  var q = query(db, query.type('post'))
  q = q.AND([query.channel("solarpunk"), query.OR([
    query.author(mix),
    query.author(mixy)
  ]))
```
Paginator keeps track of offset, limit and number of results.


arj03:
Also I'm expecting to put the query thing on jitdb so its not a seperate file

arj03:
And it would probably be easier to read if it was:

```javascript
var q = query(db, filter.type('post'))
  q = q.AND([filter.channel("solarpunk"), filter.OR([
    filter.author(mix),
    filter.author(mixy)
  ]))
```

staltz:
Brief because my wrists.

## Avoid mutating `var`

It's somewhat unusual and prone to mistakes, so we could do something like (I think LINQ does this):

```js
const q = query(db, query.type('post'))
  .AND(query.channel("solarpunk"))
  .AND(query.OR([
    query.author(mix),
    query.author(mixy)
  ]))
```

## All lowercase

From an API perspective, all uppercase `AND` shouldn't be different from `author` or `channel`, so we could have:

```js
const q = query(db, query.type('post'))
  .and(query.channel("solarpunk"))
  .and(query.or([
    query.author(mix),
    query.author(mixy)
  ]))
```

## Dot chaining or separate operators?

We have two ways to chain the operations. Either dot chaining (LINQ style) or separate operator functions (pull-stream style). There are pros/cons for each.

Dot chaining (LINQ):

```c#
var results =
     SomeCollection
        .Where(c => c.SomeProperty < 10)
        .Select(c => new {c.SomeProperty, c.OtherProperty});

results.ForEach(x => {Console.WriteLine(x.ToString());})
```

Operator functions (pull-stream):

```js
pull(
  pull.values(['file1', 'file2', 'file3']),
  pull.asyncMap(fs.stat),
  pull.collect(function (err, array) {
    console.log(array)
  })
)
```


Pros of dot chaining: somewhat shorter to type, doesn't require a pipe function wrapping everything, looks more obviously like a chain of operations.

Cons of dot chaining: as the number of available operators grows (either core operators or community driven), the library grows a lot and every third-party operator will want to add themselves to the core operator, leading to a bunch of PRs. Other time, the library becomes huge. That's been my experience designing and implementing [xstream](https://github.com/staltz/xstream/), [rxjs](https://github.com/Reactive-Extensions/RxJS/) & [ixjs](https://github.com/ReactiveX/IxJS/) (this is pretty much LINQ for JS, by the way. I know the creator of LINQ, Erik Meijer), [callbags](https://github.com/callbag/callbag/wiki).

Pros of operator functions: they can be stored in any module, and can be composed more easily without hacking the JS prototype object.

Cons of operator functions: the same as the pros for dot chaining, inverted.

**I would recommend operator functions**, especially for two reasons: (1) easier to compose third-party operators, (2) we already seem to have some anyway-needed separate functions such as `author()` and `channel()` etc.

In that light, here's an API suggestion:

```js
const q = query(
  db,
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ]))
)
```

We could also change the first line, e.g.:

```js
const q = query(
  query.from(db), // <==== this
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ]))
)
```

I have other suggestions, keep on reading.

## Clearly indicate what method "executes/consumes/delivers" the data

I noticed the way you finally "execute" the query and get data is `paginate`. I would suggest instead that the definition of pagination is an operator, and the consumption of data is another thing. Example:

```js
const q = query(
  db,
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ])),
  query.paginate(10) // just limit
)
```

If you need to set the offset where to start the pagination, we can have another operator called `startFrom`, if that operator is missing then the default offset is zero.

```js
const q = query(
  db,
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ])),
  query.startFrom(3), // offset
  query.paginate(10) // limit
)
```

**Consumption** is then defined by `toFoo` operators, e.g. `toCallback`:

```js
const q = query(
  db,
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ])),
  query.startFrom(3),
  query.paginate(10),
  query.toCallback((err, msgs) => {
    assert(msgs.length, 10) 
  })
)
```

`toCallback` bundles together all the metadata that was gathered by the chain of operators and finally executes the query, delivering results to the callback. If you need more pages, then we can use `toPullStream` (pull streams are actually exactly what we need here, not nested callbacks):

```js
const q = query(
  db,
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ])),
  query.startFrom(3),
  query.paginate(10),
  query.toPullStream()
)

pull(
  q,
  pull.drain((err, page) => {
    assert(page.length, 10) 
  })
)
```

Or getting rid of `q`:

```js
pull(
  query(
    db,
    query.type('post'),
    query.and(query.channel("solarpunk")),
    query.and(query.or([
      query.author(mix),
      query.author(mixy)
    ])),
    query.startFrom(3),
    query.paginate(10),
    query.toPullStream()
  ),
  pull.drain((err, page) => {
    assert(page.length, 10) 
  })
```

As a bonus, we can also have `toAsyncIter` which should be easy to code in max 20 lines of code or something like that:

```js
const q = query(
  db,
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ])),
  query.startFrom(3),
  query.paginate(10),
  query.toAsyncIter()
)

for await (let page of q) {
  assert(page.length, 10)
}
```

And `toPromise` as an alternative to `toCallback`:

```js
const q = query(
  db,
  query.type('post'),
  query.and(query.channel("solarpunk")),
  query.and(query.or([
    query.author(mix),
    query.author(mixy)
  ])),
  query.startFrom(3),
  query.paginate(10),
  query.toPromise()
)

const msgs = await q;
assert(msgs.length, 10)
```

## Destructuring imports

To make code even less verbose:

```js
const {query, fromDB, type, and, or, toAsyncIter} = require('jitdb/operators')
const {author, channel, startFrom, paginate} = require('jitdb-community-extras')

const q = query(
  fromDB(db),
  type('post'),
  and(channel("solarpunk")),
  and(or([ author(mix), author(mixy) ])),
  startFrom(3),
  paginate(10),
  toAsyncIter()
)

for await (let page of q) {
  assert(page.length, 10)
}
```

The specifics of which operator goes where is debatable, the code above is just showing that operators can come from different modules.

-------

What do you think?

And LOL, I said I would be brief. No wonder my wrists are in pain.

arj03:
Great, I like the syntax. This also answers the nagging pull question I had in my head. The operators in bundles is close to the filter idea that was refactored, so all for that :)

Regarding *Avoid mutating var* then in linq you can do:

`var a = db.Where(x => x.age > 10)`

and then later:

```
if (filterName)
  a = a.Where(x =>x .name.StartsWith(filterName))
```

But yes, this should *only* be for the if something case, the normal case would not need this reassigning.

staltz:
Here's an idea for the `if` case:

```js
const q = query(
  fromDB(db),
  type('post'),
  and(channel("solarpunk")),
  and(or([ author(mix), author(mixy) ])),
  weHaveOffset ?     // <-----------
    startFrom(3) : 
    null,
  paginate(10),
  toPullStream()
)
```

staltz:
By the way, I can help implement this stuff, earliest tomorrow, today I'm busy with MV release.

arj03:
Maybe we can have a shared coding session tomorrow where I type to help your wrists?

mixmix:
**1. Staltz I like your proposed api,** it feels clear to me on all points

2. For speccing / proposal threads **I highly recommend numbering proposals**, and parts on them - it makes it infinitely easier to say "I liked this, i font know about that"

**3. Do we need to roll our own here :**
a) is there an existing library that's already done this
b) could we just use SQL? (as the base layer at least, then add sugar on top) 

(3) is about saving us a lot of time and energy, as history has shown us there is A LOT involved in making dbs.

**4. How do you do JOINS with this API?**
- we are dealing with relational data, if we don't design an API which supports (now or in future builds) this i think it's a mistake
- eg find all gatherings which are being attended by one of my close friends? 

**5. How do I do backlinks with this API?
**
- eg, I want all type profile with a root %msgId
- actually I generally want the root message itself and the backing messages that are in the same tangle

arj03:
Thanks for the feedback @mixmix. It is really valuable as you are probably the most knowledgeable person about writing apps for the classic ssb database interface. One thing that I think is different is that instead of having one hammer for everything (backlinks, about) things will be a bit more granular.

> How do you do JOINS with this API?

In general, jitdb can do join based on the [DATA operator](https://github.com/arj03/jitdb#queryoperation-offset-limit-reverse-cb). This means that you would be able to do something like:

```
const mentions = sbot.db.social.getMessagesByMention(feedId) <- index outside jitdb
const posts = mentions.AND(filter.type('post')).all() <- probably the api above, but for illustration purpose of join
```

The [gathering](https://github.com/ssbc/ssb-gathering-schema) example is a good one (except it uses about ;-)) so lets try and break that down.

Close friends would be hops 1 I assume, so pseudo code:
```
const closeFriends = sbot.friends.getHops(1) <- array of ids
const isAttending = query(filter.type('about')).all().reduce(by about and attendee.link).filter(x => closeFriends.filter(cf => x.attendee.includes(cf))
```

This would give you the message ids that then would have to be hydrated.

You could replace the about type [filter](https://github.com/arj03/jitdb/blob/fbe943032257b17b253e8ad307c15f04f6cab8f6/helpers.js#L12) with something more specific. And a note on these extra indexes in jitdb, on my laptop it gives a one time hit of 8 seconds indexing + 20kb index file for 1.4 mill messages and afterwards it is fast to query.

> eg, I want all type profile with a root %msgId

```
const profiles = query(db, filter.type('profile'))
const roots = sbot.db.social.getMessagesByRoot(message.key)
const results = profiles.AND(roots).all()
```

>  I generally want the root message itself and the backlinking messages that are in the same tangle

So for latest 10 post threads this would be:

```
query(db, filter.type('post')).paginate(0, 10, (err, results) => {
  const threads = results.map(message => { message, backlinks: sbot.db.social.getMessagesByRoot(message.key) })
})
```

backlinks would be a lazy query.

mixmix:
Thanks @arj03. What's the sbot.db.social? Is that a hypothetical plugin or something which exists I should know about?

What does "backlinks would be a lazy query" mean?

staltz:
> What's the sbot.db.social? Is that a hypothetical plugin or something which exists I should know about?

Something that we'll build or are building now

arj03:
> What does "backlinks would be a lazy query" mean?

It means it won't run any db queries until you do a `query(db, backlinks).all()` on them.

staltz:
This code currently works!

```js
const {query, fromDB, and, or, type, author, toPromise} = require('jitdb/query2')

const results = await query(
  fromDB(db),
  and(type('post')),
  or(author(mix), author(mixy), author(arj)),
  toPromise()
)

console.log(results)
```

staltz:
@arj03 I believe we need to rethink `or()`, because the code above sounds like it means "all msgs of type post OR all msgs from these authors regardless of type", which is not what it's doing. Also, what if someone actually wanted that, how would the query look like? We probably need to remove the implicit "and + or" inside the current implementation of `or()`.

arj03:
Good point @staltz. Lets remove the implicit AND.

mixmix:
Agree, I would have expected

```js
const​ ​results​ ​=​ ​await​ ​query​(​
  ​fromDB​(​db​)​,​
  ​and​(​type​(​'post'​)​)​,​
  and(
  ​  or​(​author​(​mix​)​,​ ​author​(​mixy​)​,​ ​author​(​arj​)​)​,
  ), 
  ​toPromise​(​)​
​)
```

Is it going to be composable like?
Here's my assumption about what I thought you meant by that earlier :

```js
const friendsPosts = query(
  and​(​type​(​'post'​)​)​,​
  and(
  ​  or​(​author​(​mix​)​,​ ​author​(​mixy​)​,​ ​author​(​arj​)​)​,
   )
)

const source = query(
  fromDB(db), 
  friendsPosts,
  and(mention(staltz)), 
  toPullStream({ live: true }) 
) 

```


This is looking really exciting. Can't wait to catch up on call (my) tonight.

staltz:
@mixmix Yes it should compose like that (that's the goal), but currently the implementation might be bugged if you try that specifically.

staltz:
Reopening to make CI run.

staltz:
@arj03 There seems to be some probabilistic race conditions with the `GT,GTE,LT,LTE` test. Sometimes it gets `actual: 2` when `expected: 3`.

arj03:
@staltz have not seen that locally, will try to reproduce. Thanks for adding CI :-)

arj03:
wup

staltz:
I'm thinking of moving all operators that assume some "SSB conventions" to somewhere out of JITDB, e.g. `seekChannel`, `seekRoot` and even `seekType` are all things that depend on SSB message (soft) conventions, maybe they belong to `ssb-db2/social`? Anyway, I'll just mark with a `FIXME` that these operators should be moved, and then eventually we can find the correct module for them.

staltz:
@arj03 I think this is done. I thought about also adding support for `offset` and `reverse` to `getAll`, but this could be done in a separate PR since #6 is already quite big.

arj03:
Great! Looking forward to reviewing, will get to that this afternoon.

arj03:
This is really great @staltz . I made some minor changes I think the only thing missing before merging is to include some examples in the README of the new query interface. Could just be taken from example more or less.

staltz:
Oh yeah, the README. Will do.

staltz:
As I write the README, I realize we probably should support this:

```js
// get all msgs
query(
  fromDB(jitdb),
  toCallback((err, msgs) => {
    console.log(msgs.length)
  })
)
```

staltz:
Ready!

arj03:
Brilliant! Thank you so much, this new api is a huge step forward.

PR #7, opened 2020-11-11
Fix tests

arj03:
I can haz CI?


arj03:
Ah lol, its a bug in the tests, it assumes the timestamps are always increasing :(

```
Float64Array [
  1605081485381,
  1605081485381,
  1605081485382,
  1605081485382
```

Issue #11, opened 2020-11-11
Move opinionated operators elsewhere


staltz:
We should keep JITDB unopinionated, JITDB should only assume: 

- There is an `async-flumelog` using `flumecodec/json`
- JSON Messages stored in the log are of the following format

```
{
  key: MsgId,
  value: {
    previous: MsgId,
    author: FeedId,
    sequence: number,
    timestamp: number,
    hash: 'sha256',
    content: any, 
    signature: string,
  };
  timestamp: number,
}
```

I.e. JITDB doesn't assume what `msg.value.content` internally looks like. Currently in JITDB we have some things that assume SSB message conventions, and these should be moved to a more opinionated module:

- [ ] `jitdb/operators/type`
- [ ] `jitdb/operators/channel`
- [ ] `jitdb/operators/isRoot`
- [ ] `jitdb/operators/isPrivate`

staltz:
I think I figured out what to do about this issue! Turns out that operators like `type`, `author`, `channel` are not the building blocks. JITDB should have operators for the *operations* that JITDB supports: DATA, EQUAL, AND, OR, GT, GTE, LT, LTE.

Currently we have some but are missing some others:

- [x] `and` for AND
- [x] `or` for AND
- [ ] `equal` for EQUAL
- [ ] `gt` for GT
- [ ] `gte` for GTE
- [ ] `lt` for LT
- [ ] `lte` for LTE
- [ ] `data` for `DATA` (I'd love to rename this to `POSITION` actually, I find the word "DATA" not descriptive and already overused for other parts inside JITDB)

`equal` wouldn't be so ergonomic to use, because it requires an `indexType` and a `seek` function (which should be written by hand intelligently to optimize for performance). That said, we could have two operators for EQUAL:

- `equal(helpers.seekType, 'post', 'type')`
- `slowEqual('value.content.type', 'post')`

`slowEqual` make it easy to quickly experiment with queries, it builds the seek function by doing a `.split('.')` on the first argument. The explicit name "slow" encourages developers to not use this if they care about performance.

In practice, opinionated operators like `type`, `author`, `channel`, `isRoot`, `isPrivate` would live in ssb-db2/social and would **wrap** `equal` so to guarantee good performance.

arj03:
[...]

PR #12, opened 2020-11-11
add operators equal / slowEqual / gt / gte / lt / lte

staltz:
So far a draft. For issue #11

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -60,6 +60,46 @@ function equal(seek, value, indexType) {
   };
 }
 
+function gt(value, indexName) {
+  return {
+    type: 'GT',
+    data: {
+      value,
+      indexName,
+    },
+  };
+}
+
+function gte(value, indexName) {
+  return {
+    type: 'GTE',
+    data: {
+      value,
+      indexName,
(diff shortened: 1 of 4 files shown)


operators.js
```diff
@@ -16,6 +17,42 @@ function toBuffer(value) {
   return Buffer.isBuffer(value) ? value : Buffer.from(value)
 }
 
+function seekFromDesc(desc) {
+  return buffer => {
+    const keys = desc.split('.')
```
arj03, line 22: This line can be moved outside inline function

arj03: Seems to make the example query go from 15.3s to 14.5s :-)

staltz: I can easily move it, but what do you think if we intentionally keep it slow, so that for production, developers feel motivated to have `equal` and never `slowEqual`?

arj03: I think its good like this. Makes it easy as you say to quickly test something.

arj03:
Great, thanks! Merged this in now and made the tiny move 1 line.

PR #17, opened 2020-11-12
ascending by default, getAll supports offset and descending

staltz:
- `getTop` and `getAll` both do ascending order by default
- `getAll` supports arguments `offset` and `descending`
- `getTop` and `getAll` implementations looked so similar, I made a common function called `getValuesFromBitsetSlice`
- Removed the optionality of the `reverse` argument
  - It must always be provided
  - Assuming we won't use `getTop`/`getAll` as a direct API, but instead we'll use higher level APIs, I wanted to reduce some of the magic and code size of the implementation
- Add test for `startFrom`+`toCallback`
- Add test for `descending`+`toCallback`

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,12 +109,12 @@ query(
 
 #### Pagination
 
-If you use `toCallback`, it gives you all results in one go. If you want to get results in batches, you should use **`toPullStream`**, **`paginate`**, and optionally `startFrom` and `ascending`.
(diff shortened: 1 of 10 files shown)


arj03:
The last fail is rather strange. Just removed a comment, must be some concurrency thing again ;-)

arj03:
Compared the performance of getAll existing with this new sorted and there is almost no overhead.

Changed the docs a bit, I wonder if we should let all return the same format as paginate. Might be nice with the duration. Lets do that in another PR maybe.

Need to check the test error but otherwise good to merge. Really like how you simplified all and paginate and used better names for variables!

arj03:
Boom :smiley:

PR #18, opened 2020-11-12
refactor index.js: rename db to log

staltz:
Tiny refactor. I was sometimes confused that this was called `db` since it's a flumelog instance, and the object `index.js` returns is later in tests called "db" (jitdb instance) too.


arj03:
Yep, much nicer. Thanks

PR #21, opened 2020-11-12
Add benchmark in ci

staltz:
Checks that duration of index building does not go over a threshold.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -10,7 +10,8 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  test:
+    name: Tests
 
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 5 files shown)


arj03:
I like the spirit of this. On my machine I get:

```
# tests 13
# pass  9
# fail  4
```
Some of the tests takes a bit longer than the max time, they appear to run fine on the CI machines. I wonder if we might run into random failures based on this. Time will tell I guess. 

Does the CI need to download the data on every commit? Also is there any kind of limit on this kind of thing?

staltz:
Good questions! I'm open to suggestions how to improve this, right now I admit it's naive and simple.

On my computer the durations are smaller than on CI, I'm surprised on your computer the numbers are greater than on CI. You used the same datasets right?

About download limits: I'll take a look. We can also use the sloop fixture which is 10x smaller. And there may be a cache functionality in CI that we have to enable.

arj03:
Another option would be to keep this here while things are under heavy development and then merge once things settle down. We can just merge into this branch from master once in a while and see how things are going.

staltz:
I'm thinking of just using the sloop fixture instead of catamaran. That should be much lighter. And GitHub downloading 4MB from GitHub itself shouldn't be bad for them. We can run the benchmark X times and take an average or a min.

staltz:
Made obsolete by #60

PR #26, opened 2020-11-13
Fix debug

arj03:
This makes debug nicer to read by converting buffer to string. Also removes the extra mkdirp. There is an extra commit in here because I didn't push, but should be meaningless.


staltz:
Beautiful, would merge, but... why remove mkdirp from package.json? It's used in tests.

arj03:
it was both in dev and main

staltz:
Ah right right, got it

PR #28, opened 2020-11-13
Add the deferred operator

arj03:


staltz:
Later I'll submit a PR to ssb-db2 that makes use of this. It actually requires the db2 instance to be attached to meta, like `meta.db2`

staltz:
@arj03 What about this PR?

operators.js
```diff
@@ -195,45 +205,83 @@ function paginate(pageSize) {
...
 }
 
+async function executeDeferredOps(ops, meta) {
+  // Collect all deferred tasks and their object-traversal paths
+  const allDeferred = [];
+  traverse(ops).forEach(function (val) {
+    if (!val) return;
+    // this.block() means don't traverse inside these, they won't have DEFERRED
+    if (this.path.length === 1 && this.key === 'meta') return this.block()
+    if (val.type === 'Buffer' && Array.isArray(val.data)) return this.block();
```
arj03, line 215: Can we make this more general maybe? The problem is if someone adds a new operator they will most likely forget to add it here.

arj03:
@staltz just have the one comment, really excited about the options this opens up. Also its nice that this doesn't touch the core.

operators.js
staltz: Apart from the Buffer check (which I think we should keep), I could also do the inverse logic: look only for AND, OR and DEFERRED. I think that's enough to look for the all the depths of the object?

staltz: Oh and by the way, if someone forgets to update this list when a new operation is added, everything will still just work. This list is just an optimization. We could remove all of these checks and it would still work. I just wanted to remove some unnecessary calculations, even though they aren't super heavy calculations.

arj03: Yeah only looking for those I think would be better.

staltz:
Updated this PR

arj03:
Brilliant! Merging and publishing a new version.

PR #34, opened 2020-11-18
replace dep pify with promisify-4loc

staltz:
I found `pify` too big. I was gonna use it in db2 as well. so instead i took a quick detour and made https://github.com/staltz/promisify-4loc (mostly copied from https://github.com/staltz/promisify-tuple)


arj03:
Looks good to me

PR #36, opened 2020-11-19
support deferred operator to return falsy values

staltz:
This is in preparation of putting `onDrain` implicitly inside `query` in db2.

If `deferred`'s callback gives a falsy value, we make an empty object instead. This is mostly to support use cases that "wait for something" before executing the query.

```js
query(
  fromDB(jitdb),
  deferred(waitForThisStuffToHappenFirst),
  and(type('post'), author(me),
  toCallback(done)
)
```


arj03:
Looks good, can be hard to judge in isolation but I trust you :)

PR #37, opened 2020-11-20
Rewrite live query single to be a more general live query

arj03:
This is still a draft as it needs operator support.


staltz:
@arj03 Looking good, let me know if/when you'd like to pair program.

I have two comments so far:

- If the API takes a callback, it's tricky to ask for just 1 live value (e.g. `pull.take(1)` style), or in general tricky to cancel (or removeListener) the async delivery of values. If the actual underlying async pipeline is a push-stream, can we just use push-streams, and then in `toCallback`/`toPromise`/`toPullStream`/`toAsyncIter` we handle each case? E.g. in the `toCallback` case, we would just take the push-stream source given by `live()` and put a `take(1)` on it so to get just one value. If people need multiple live values, they inherently need either `toPullStream` or `toAsyncIter`
- I got confused by the naming `deferred`, I thought it was `type: 'DEFERRED'` op. Could we aim for each word to have primarily one meaning? Maybe this one could be named "positional"? Or just "offset" (if you're not intending that `live` supports SEQS)

arj03:
@staltz would love to pair with you on the operator part. I'm chasing down a strange bug atm, need that fixed first ;-)

arj03:
Going to close this and open another PR where the history is squashed

staltz:
Can we delete branch `live`?

arj03:
done

PR #38, opened 2020-11-25
Rewrite live query single to be a more general live query

arj03:
This is a continuation of #37 where live has been rewritten to return a pull stream instead of individual values. A pull stream makes it easier to reason about things like cancelling.

This PR also includes a fix for live on an empty database. We could do that in a separate PR if better.


staltz:
My thoughts so far `db.live(op, (err, pullstream) => { ... })` could be instead `const pullstream = db.live(op)`, because this will greatly simplify `./operators.js#toPullStream`, as well as simplify writing tests. Also, the `cb` is never called with error in the 1st arg. If you want, I can help refactor this.

arj03:
Right, would be a good idea. Would be nice if you can do that.

staltz:
I'll do it as a PR from `mybranch` to `live-pully` so you get to review it before it pops into here.

arj03:
Tests works after merge @staltz

staltz (approved):
Looks all good now. I would maybe change some comments and variable names to avoid confusion around "deferred" (I would like this name to refer to only one thing), but that can be done in another PR and it's just code quality related

arj03:
Wup, thanks for the help on this @staltz :-)

PR #39, opened 2020-11-25
Refactor db.live() to return pull-stream

staltz:
Does the refactor I mentioned here https://github.com/ssb-ngi-pointer/jitdb/pull/38#issuecomment-733643335 and also adds support for `live` in `toAsyncIter()`


staltz:
Funny that the tests don't run for this PR, but I have run the tests locally and they pass.

arj03:
They also work for me. I don't see any tests being run on this repo. The changes looks good. Nice refactor as well. A was thinking of using pull-cont but I see pull-async does the same thing.

PR #42, opened 2020-11-25
refactor index.js a bit

staltz:
@arj03 I'm getting familiar with the codebase but simultaneously I keep getting lost because of the difficulty to navigate the file (e.g. names that mean something else or big functions inside big functions) so I'll begin refactor this. I'll do it incrementally so each refactor PR is easy to review.

this time I did:

- rename `setupIndex` to `sanitizeOpData` because the index itself is actually not touched, it's the `op.data` fields that are prepared
- exported methods are moved out of the object notation, so the resulting exported object is small and easy to glance

No actual logic was harmed in this process

### DIFF
--- a/index.js
+++ b/index.js
@@ -512,7 +512,7 @@ module.exports = function (log, indexesPath) {
     )
   }
 
-  function setupIndex(op) {
+  function sanitizeOpData(op) {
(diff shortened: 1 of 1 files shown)


arj03:
Thanks

PR #43, opened 2020-11-26
Refactor2

staltz:
- move functions up or down (this is in preparation of putting them in other files)
- remove lodash.debounce
- one variable `buffer` is `data.value` but the function used `data.value`, so I just replaced it with `buffer`


package.json
```diff
@@ -14,7 +14,6 @@
     "debug": "^4.2.0",
     "idb-kv-store": "^4.5.0",
     "jsesc": "^3.0.2",
-    "lodash.debounce": "^4.0.8",
```
arj03, line 17: nice catch

PR #44, opened 2020-11-26
refactor: rename all 'data' variables

staltz:
Ok, this one is a real refactor that I'm open to discuss and iterate if you wish.

First I noticed that there are two very different types of indexes:

- TypedArray indexes: offset, timestamp, sequence
- TypedFastBitSet indexes: all others created JIT

It took me a while to realize that the first one `index.data` is a TypedArray, but the second one `index.data` is a bitset datastructure that has `index.data.count` and `index.data.words`. The fact that `index.data.words` was also a TypedArray made this confusing because at some point I thought "how can there be a TypedArray inside a TypedArray?" but the truth was that `index.data.words` meant that `index.data` was a bitset.

So one thing i did was rename these two types of indexes. The first type is a "core index" and the second type is a (regular) "index". They are saved to disk differently too.

As you can see, there is a lot of the name `data` used everywhere for different meanings:

- index.data as a TypedFastBitset
- index.data as a TypedArray
- op.data
- each response to a log.stream drain
- a buffer
- an index
- the results of a paginate call

I looked at using more descriptive names for those, and I came up with this:

- index.data as a TypedFastBitset
  - **index.bitset**
- index.data as a TypedArray
  - **index.tarr** (I considered `typedArray` but that's a bit too long, `tarr` is nice that it looks like `arr`. any resemblance with Dominic Tarr is just a happy coincidence that we can keep as a tribute to him :)
- op.data
  - I kept this one as is
- each response to a log.stream drain
  - **record**, that's what async-flumelog docs say they are, seemed suitable
- a buffer
  - **buf** or **buffer**
- an index
  - **index** or **idx**
- the results of a paginate call
  - **results**

This means that you can grep the codebase for `data` and you'll basically only get `op.data`. You can grep for `record` and it will mean that one thing, and so forth.

Looking forward to hear what you think

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,13 +33,13 @@ module.exports = function (log, indexesPath) {
     return sanitize(result)
   }
 
-  function saveTypedArray(name, seq, count, arr, cb) {
+  function saveTypedArray(name, seq, count, tarr, cb) {
     const filename = path.join(indexesPath, name + '.index')
     if (!cb) cb = () => {}
 
-    debug('writing index to', filename)
+    debug('writing index to %s', filename)
 
-    const dataBuffer = toBuffer(arr)
+    const dataBuffer = toBuffer(tarr)
     var b = Buffer.alloc(8 + dataBuffer.length)
     b.writeInt32LE(seq, 0)
     b.writeInt32LE(count, 4)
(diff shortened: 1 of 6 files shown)


arj03:
I literally lol'ed when I read the tarr name. Lets keep that :-)

I really love your attention to detail here. Overall I feel this is much more readable, even as the author of most of the original code. The only thing I would change is maybe the name answer, but now I'm nitpicking. Great work!

PR #45, opened 2020-11-27
refactor: extract files.js from index.js

staltz:
This one is a lot of shuffling. I wanted to move persistence logic to its own file, so now `atomic-file` logic and low-level FS things are in a file called `files.js`, it basically handles just save/load TypedArrays or save/load Bitsets.

`files.js` has no notion of what is an "index", so as I did this refactor I made other parts of the code simpler, see e.g. `saveIndex(name, index)` and the updated tests.

### DIFF
--- a/files.js
+++ b/files.js
@@ -1,0 +1,98 @@
+const jsesc = require('jsesc')
+const sanitize = require('sanitize-filename')
+const TypedFastBitSet = require('typedfastbitset')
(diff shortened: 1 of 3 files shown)


arj03:
Great! I actually needed exactly the save and load part of private messages in db2, so this refactor is very welcome :-)

PR #46, opened 2020-11-29
refactor: rename vars and functions related to syncing

staltz:
Stuff such as:

- var => let or const
- triple equality preferred over double equality
  - in some cases (where the two types don't match and coercion is needed), triple is faster
- renamed "values" to "messages" in some cases
  - "values" is an overloaded name that can mean ~3 things, so I renamed one of the cases to "messages" when the output is the BIPF-decoded record
  - this makes the word "messages" in README and tests consistent with what is in the code too
- pulled out some nested functions when there was no strict need for the nesting
- got rid of `ensureOffsetIndexSync` which is now handled by `ensureIndexSync`
- some more renames

PS: I'm starting to understand the codebase and I have a few more refactors in mind, one file extraction and another is about `seq` and `offset` and I think their meaning is flipped around sometimes

### DIFF
--- a/index.js
+++ b/index.js
@@ -373,12 +373,13 @@ module.exports = function (log, indexesPath) {
         : Buffer.from(op.data.value)
   }
 
-  function ensureIndexSync(opOrName, cb) {
-    const isName = typeof opOrName === 'string'
-    const op = isName ? { data: { indexName: opOrName } } : opOrName
-    const indexName = isName ? opOrName : op.data.indexName
-    if (log.since.value > indexes[indexName].seq) updateIndex(op, cb)
+  function ensureIndexSync(op, cb) {
+    if (log.since.value > indexes[op.data.indexName].seq) updateIndex(op, cb)
     else cb()
+  }
+
+  function ensureOffsetIndexSync(cb) {
+    ensureIndexSync({ data: { indexName: 'offset' } }, cb)
   }
 
   function filterIndex(op, filterCheck, cb) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -711,6 +715,7 @@ module.exports = function (log, indexesPath) {
     )
   }
 
+  // FIXME: why do we need to export this? It's simple and only used in 1 test
```
arj03, line 718: This goes together with querySeq. I think those two methods were something that I added in the beginning thinking they would be a good idea and now I'm not sure sure. I checked and never really used them, so I'd actually be in favour of just removing them.

index.js
```diff
@@ -373,25 +373,115 @@ module.exports = function (log, indexesPath) {
         : Buffer.from(op.data.value)
   }
 
-  function ensureOffsetIndexSync(cb) {
-    if (log.since.value > indexes['offset'].seq)
-      updateIndex({ data: { indexName: 'offset' } }, cb)
+  function ensureIndexSync(opOrName, cb) {
```
arj03, line 376: I think I would prefer this to be ensureIndexSync as it was before, and then a helper function to generate an "fake" op from name.

arj03: the function was a bit hard for me to read. It looks as if it does a bunch of things, but mainly just mangles stuff into the right shape

arj03:
Overall this looks great. Much better function names. Would like that ensureIndexSync tweaked a bit before merging, and the querySeq can be done in another commit.

index.js
staltz: Fine by me, I'll do that

index.js
staltz: Oh, deleting code always feels good. Will do

arj03: It was there because I thought we might need it for about and contacts, but I'm not sure sure, we can always add it again if its just a simple PR that removes it.

arj03:
Perfecto. Thanks

PR #47, opened 2020-11-29
remove querySeq() and getSeq()

staltz:
See https://github.com/ssb-ngi-pointer/jitdb/pull/46#discussion_r532245044 for context


arj03:
Great, removed from README as well

PR #48, opened 2020-11-29
fix corner cases of executing deferred ops

staltz:
Some small details in the logic seemed wrong (i.e. DEFERRED inside AND or OR might be bugged).


staltz:
sorry, was missing tests. added now

arj03:
Great catch!

PR #51, opened 2020-12-01
Replace atomic-file with atomically-universal

arj03:
As the header says. Atomically should have much better error handling when running on node.

Side note: i find the auto format of promises wierd. Looks like things are floating. Anyway minor point.

### DIFF
--- a/files.js
+++ b/files.js
@@ -1,7 +1,7 @@
 const jsesc = require('jsesc')
 const sanitize = require('sanitize-filename')
 const TypedFastBitSet = require('typedfastbitset')
-const AtomicFile = require('atomic-file/buffer')
+const { readFile, writeFile } = require('atomically-universal')
 const toBuffer = require('typedarray-to-buffer')
 
 function saveTypedArrayFile(filename, seq, count, tarr, cb) {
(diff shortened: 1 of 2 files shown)


files.js
```diff
@@ -13,29 +13,27 @@ function saveTypedArrayFile(filename, seq, count, tarr, cb) {
   b.writeInt32LE(count, 4)
   dataBuffer.copy(b, 8)
 
-  const f = AtomicFile(filename)
-  f.set(b, cb)
+  writeFile(filename, b).then(cb).catch(cb)
```
staltz, line 16: Are you sure this promise resolves with undefined or null? If it resolves with anything else, then it would do `.then(x => cb(x))` which means delivering `x` as an error, which is probably not what we want

staltz:
Just one comment, other than that, looks ready to merge

files.js
arj03: Good spotting. I did check that in both node and browser this returns undefined, but in order to be sure that this will always work I made a small change so it will never return anything. Thanks

mixmix:
@arj03 you might like `prettier-prettier` or `semi-prettier`

Issue #40, opened 2020-11-25
Radix idea: index that can look up any mentioned msg key


staltz:
some more crazy ideas to use bitvectors: I know it's not feasible to have jitdb indexes for votes, because there are two many vote roots (thus one index for each vote root), and so currenly in ssb-db2 we use leveldb for that. But what if we did something similar to radix sort? 

Lets say we take the msgid at `msg.value.content.vote.link`, convert it to buffer, and take the 1st bit, then all messages that have the 1st bit of `msg.value.content.vote.link` equals 1 are indexed in a bitvector. You could do that for the 2nd and 3rd bits too, with the goal being to just limit the search space, and then once it's small enough, just scan the results in memory. This could be combined with #8.

i.e. these bitvectors:
- msg_value_content_vote_link_bit1_is_1.index (**realistic approx. size: 160KB**)
- msg_value_content_vote_link_bit2_is_1.index
- msg_value_content_vote_link_bit3_is_1.index

Then we can create a query such as `votesFor(msgid)`, pick the first 3 bits of `msgid` and then AND queries for the corresponding bitvectors mentioned in the list above. Maybe an optimal amount of these indexes could be about ~16?

If we have 16 bitvectors for each shape of mentioned key, then we start to have a lot of indexes, e.g.:

- msg_key_bit1.index
- msg_key_bit2.index
- ...
- msg_key_bit16.index
- msg_value_content_mentions_feedid_bit1.index
- msg_value_content_mentions_feedid_bit2.index
- ...
- msg_value_content_mentions_feedid_bit16.index
- msg_value_content_mentions_msgid_bit1.index
- msg_value_content_mentions_msgid_bit2.index
- ...
- msg_value_content_mentions_msgid_bit16.index
- msg_value_content_vote_link_bit1.index
- msg_value_content_vote_link_bit2.index
- ...
- msg_value_content_vote_link_bit16.index

In that case, maybe we could save all 16 vectors into one file, then we'd have bitmatrix files. Nx16 as opposed to Nx1:

- msg_key_bitmatrix.index (**realistic approx. size: 2.5MB**)
- msg_value_content_mentions_feedid_bitmatrix.index
- msg_value_content_mentions_msgid_bitmatrix.index
- msg_value_content_vote_link_bitmatrix.index

### How much does 16 bits reduce the search space?

[...]

PR #52, opened 2020-12-02
prefix indexes

staltz:
For issue #40, this is a work in progress but most of it is working.

- [x] Experiments with 8 bit, 16 bit, 32 bit
- [x] Working implementation
- [x] Benchmarks
- [x] Consider the API
- [x] Tests
- [x] Clean up code

Note about the API: I think for a long time what would be a good operator name. But it turns out that semantically this is the same as `equal`, it's just a different way of doing `equal`, so I opted for the same name, but indicating that it's using prefix indexes `equalViaPrefix`. I would maybe prefer another API: `equal(seeker, value, opts)` where `opts: {indexType, indexAll, prefix}` so to use this new algorithm you would write `equal(seekVoteLink, msgid, {prefix:true})` or `slowEqual('value.content.vote.link', msgid, {prefix:true})`

Random notes about early perf benchmarks:

```
and(blog, or(author1, author2))

  - 3 records found
  - duration when cold everything missing: 6900ms
  - duration when cold 3 missing indexes:  3000ms
  - duration when cold 2 missing indexes:  2100ms
  - duration when cold 1 missing index:    2000ms
  - duration when warm query:                20ms

votes.link=msgid (prefix 16 bits from base64 decoded)

  - 24 records found
  - duration when cold everything missing:   5300ms
  - duration when cold prefix index missing: 4000ms
  - duration when warm query:                  50ms

slowEqualViaPrefix votes.link=msgid (prefix 32 bits from base64 as is in log.bipf)

  - 17 records found, some are false positives
  - duration when cold everything missing:   4800ms
  - duration when cold prefix index missing: 3300ms
  - duration when warm query:                  40ms

equalViaPrefix votes.link=msgid (prefix 32 bits from base64 as is in log.bipf)

  - 17 records found, some are false positives
  - duration when cold everything missing:   3600ms
  - duration when cold prefix index missing: 2000ms
  - duration when warm query:                  30ms

equalViaPrefix votes.link=msgid (prefix 32 bits from base64 as is in log.bipf)
with additional iteration over the results to remove false positives

  - 15 records found
  - duration when cold everything missing:   4600ms
  - duration when cold prefix index missing: 2100ms
  - duration when warm query:                  40ms
```

The prefix index file is the same size as the `offset.index` and `sequence.index`, roughly ~5MB for a production dataset.

### DIFF
--- a/index.js
+++ b/index.js
@@ -495,7 +495,7 @@ module.exports = function (log, indexesPath) {
     const target = op.data.value
     const targetPrefix = target.readUInt32LE(0)
     const prefixIndex = indexes[op.data.indexName]
-    for (let i = 0, len = prefixIndex.tarr.length; i < len; ++i) {
+    for (let i = 0; i < prefixIndex.count; ++i) {
       if (prefixIndex.tarr[i] === targetPrefix) bitset.add(i)
     }
     const getRec = promisify(getRecord)
--- a/index.js
+++ b/index.js
@@ -241,8 +241,7 @@ module.exports = function (log, indexesPath) {
   }
 
   function updateIndexValue(opData, index, buffer, offset) {
-    const status = checkValue(opData, buffer)
-    if (status) index.bitset.add(offset)
+    if (checkValue(opData, buffer)) index.bitset.add(offset)
   }
 
   function updateAllIndexValue(opData, newIndexes, buffer, offset) {
--- a/index.js
+++ b/index.js
@@ -493,7 +493,7 @@ module.exports = function (log, indexesPath) {
     })
   }
 
-  async function matchAgainstPrefix(op, cb) {
+  function matchAgainstPrefix(op, cb) {
     const bitset = new TypedFastBitSet()
     const target = op.data.value
     const targetPrefix = target.readUInt32LE(0)
--- a/index.js
+++ b/index.js
@@ -588,14 +588,14 @@ module.exports = function (log, indexesPath) {
   }
 
   function executeOperation(operation, cb) {
-    const orphanOps = []
+    const opsMissingIndexes = []
     const lazyIndexes = []
 
     function detectMissingAndLazyIndexes(ops) {
       ops.forEach((op) => {
         if (op.type === 'EQUAL') {
           sanitizeOpData(op)
-          if (!indexes[op.data.indexName]) orphanOps.push(op)
+          if (!indexes[op.data.indexName]) opsMissingIndexes.push(op)
           else if (indexes[op.data.indexName].lazy)
             lazyIndexes.push(op.data.indexName)
         } else if (op.type === 'AND' || op.type === 'OR')
@@ -614,13 +614,15 @@ module.exports = function (log, indexesPath) {
     }
 
     function createMissingIndexes() {
-      if (orphanOps.length > 0) createIndexes(orphanOps, getBitset)
+      if (opsMissingIndexes.length > 0)
+        createIndexes(opsMissingIndexes, getBitset)
       else getBitset()
     }
 
     detectMissingAndLazyIndexes([operation])
 
-    if (orphanOps.length > 0) debug('missing indexes: %o', orphanOps)
+    if (opsMissingIndexes.length > 0)
+      debug('missing indexes: %o', opsMissingIndexes)
 
     if (lazyIndexes.length > 0)
       loadLazyIndexes(lazyIndexes, createMissingIndexes)
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@ const pull = require('pull-stream')
 const toPull = require('push-stream-to-pull-stream')
 const pullAsync = require('pull-async')
 const TypedFastBitSet = require('typedfastbitset')
+const promisify = require('promisify-4loc')
 const bsb = require('binary-search-bounds')
 const debug = require('debug')('jitdb')
 const {
@@ -97,6 +98,17 @@ module.exports = function (log, indexesPath) {
                 cb()
               }
             )
+          } else if (file.endsWith('.32prefix')) {
+            // Don't load it yet, just tag it `lazy`
+            indexes[indexName] = {
+              seq: -1,
+              count: 0,
+              tarr: new Uint32Array(16 * 1000),
+              lazy: true,
+              prefix: 32,
+              filepath: path.join(indexesPath, file),
+            }
+            cb()
           } else if (file.endsWith('.index')) {
             // Don't load it yet, just tag it `lazy`
             indexes[indexName] = {
@@ -133,17 +145,29 @@ module.exports = function (log, indexesPath) {
     saveBitsetFile(filename, index.seq, index.bitset, cb)
   }
 
-  function growCoreIndex(coreIndex, Type) {
-    debug('growing index')
-    const newArray = new Type(coreIndex.tarr.length * 2)
-    newArray.set(coreIndex.tarr)
-    coreIndex.tarr = newArray
+  function savePrefixIndex(name, prefixIndex, count, cb) {
+    debug('saving prefix index: %s', name)
+    const num = prefixIndex.prefix
+    const filename = path.join(indexesPath, name + `.${num}prefix`)
+    saveTypedArrayFile(filename, prefixIndex.seq, count, prefixIndex.tarr, cb)
+  }
+
+  function growTarrIndex(index, Type) {
+    let newLength = index.tarr.length * 2
+    // FIXME: even better would be to bring back the exponential jumps, but
+    // when reaching the end of the log stream, cut off all the empty space
+    const diff = Math.min(newLength - index.tarr.length, 64000)
+    newLength = index.tarr.length + diff
+    debug('growing index by +%d', diff)
+    const newArray = new Type(newLength)
+    newArray.set(index.tarr)
+    index.tarr = newArray
   }
 
   function updateOffsetIndex(offset, seq) {
     if (offset > indexes['offset'].count - 1) {
       if (offset > indexes['offset'].tarr.length)
-        growCoreIndex(indexes['offset'], Uint32Array)
+        growTarrIndex(indexes['offset'], Uint32Array)
 
       indexes['offset'].seq = seq
       indexes['offset'].tarr[offset] = seq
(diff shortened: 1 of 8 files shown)


arj03:
First of all great work on this. I like the concept overall. It's pretty crazy that its so quick to loop over a million values, modern processors are really optimized for loops :-)

As for API, I like this one: `equal(seeker, value, opts)`

For db2 indexes, I could see this used for vote + root links in social. But I guess not mentions? And for base it would be interesting to try this for key index as that is probably quite big in base. And this is going to really interesting to test for about and friends.

I have similar query numbers:

Vote query: 35ms
Vote query db2: 26ms

I'll do a review of the code now.

index.js
```diff
@@ -133,17 +145,29 @@ module.exports = function (log, indexesPath) {
...
+  function savePrefixIndex(name, prefixIndex, count, cb) {
+    debug('saving prefix index: %s', name)
+    const num = prefixIndex.prefix
+    const filename = path.join(indexesPath, name + `.${num}prefix`)
+    saveTypedArrayFile(filename, prefixIndex.seq, count, prefixIndex.tarr, cb)
+  }
+
+  function growTarrIndex(index, Type) {
+    let newLength = index.tarr.length * 2
+    // FIXME: even better would be to bring back the exponential jumps, but
```
arj03, line 157: Yes, we really need to exponential back here. There is quite a big perf diff. Maybe with a trim at save function.

index.js
```diff
@@ -199,11 +223,26 @@ module.exports = function (log, indexesPath) {
...
+      index.seq = seq
+      index.count = offset + 1
+      return true
+    }
+  }
+
   function updateIndexValue(opData, index, buffer, offset) {
     const status = checkValue(opData, buffer)
     if (status) index.bitset.add(offset)
-
```
arj03, line 205: Nice, even better make this function a 1 line

index.js
```diff
@@ -274,20 +316,31 @@ module.exports = function (log, indexesPath) {
...
+          else saveIndex(op.data.indexName, index)
+        }
 
         cb()
       },
     })
   }
 
-  function createIndexes(missingIndexes, cb) {
+  function createIndexes(orphanOps, cb) {
```
arj03, line 329: I actually like the missingIndexes name better. I'm not exactly sure what orphan means here.

staltz: Oh, let me explain (and I'm open to changing it back): I general take the name to indicate the type of the thing, and in this case `missingIndexes` was not an array of indexes, it was an array of `op.data`. This confused me for a while, and changing the name helped me to know that these were actually ops, and they were passed to other functions like `updateIndexValue(opData, .....)`

staltz:
> First of all great work on this. I like the concept overall.

Thanks!

> I have similar query numbers:
>
> Vote query: 35ms
> Vote query db2: 26ms

Those are using leveldb? Do you think that 40ms with prefix indexes is acceptable or not acceptable?

> For db2 indexes, I could see this used for vote + root links in social. But I guess not mentions? And for base it would be interesting to try this for key index as that is probably quite big in base. And this is going to really interesting to test for about and friends.

Yeah I'm not sure how to do mentions, we could perhaps support the first mention in the mentions array, but not all of them, unless we do a prefix index for the first mention, another prefix index for the second mention, the third mention, etc, but that seems wasteful.

arj03:
> > I have similar query numbers:
> > Vote query: 35ms
> > Vote query db2: 26ms
> 
> Those are using leveldb? Do you think that 40ms with prefix indexes is acceptable or not acceptable?

I think we should weight this for the different scenarios. It might be possible to make this a bit faster as well.

> > For db2 indexes, I could see this used for vote + root links in social. But I guess not mentions? And for base it would be interesting to try this for key index as that is probably quite big in base. And this is going to really interesting to test for about and friends.
> 
> Yeah I'm not sure how to do mentions, we could perhaps support the first mention in the mentions array, but not all of them, unless we do a prefix index for the first mention, another prefix index for the second mention, the third mention, etc, but that seems wasteful.

I have a hunch that we might make a compound prefix for these that would have less resolution but in practice might work out. Lets iterate on that idea after this. It's not really related, was just thinking ahead :-)

index.js
arj03: How about opsNotIndexed instead then?

index.js
staltz: So, I'm not sure what to do here, but I'll share some numbers:

- exponential jumps unlimited
  - `and(blog, or(author1, author2))` with every index missing: 4400ms
  - `and(blog, or(author1, author2))` with non-core indexes missing: 3000ms
  - `and(blog, or(author1, author2))` warm query: 18ms
  - `equal(seekVoteLink, msgid, prefix)` with every index missing: 3400ms
  - `equal(seekVoteLink, msgid, prefix)` with non-core indexes missing: 2000ms
  - `equal(seekVoteLink, msgid, prefix)` warm query: 45ms
- exponential jumps limited to 64000
  - `and(blog, or(author1, author2))` with every index missing: 4600ms
  - `and(blog, or(author1, author2))` with non-core indexes missing: 3000ms
  - `and(blog, or(author1, author2))` warm query: 18ms
  - `equal(seekVoteLink, msgid, prefix)` with every index missing: 3500ms
  - `equal(seekVoteLink, msgid, prefix)` with non-core indexes missing: 2100ms
  - `equal(seekVoteLink, msgid, prefix)` warm query: 40ms

So there's a small CPU perf difference, but more significant storage difference.

staltz:
> It might be possible to make this a bit faster as well.

Yep, I was excited to let you review this and see if there's anything smarter I can do. 

Also, I forgot now what the leveldb vote index in ssb-db2 looked like in performance for the cold case, where it needs to build the leveldb before answering the query. Do you remember roughly what it was?

> I have a hunch that we might make a compound prefix for these that would have less resolution but in practice might work out. Lets iterate on that idea after this. It's not really related, was just thinking ahead :-)

:100: for "low resolution" strategies

staltz:
Idea: compression of the prefix index files might make a huge difference (I suppose primarily in storage? Maybe CPU diference too, although I'm not sure how)

arj03:
> Yep, I was excited to let you review this and see if there's anything smarter I can do.

I'll give it a go a bit later today or tomorrow. It shouldn't change anything here and I think we could even merge this without.
 
> Also, I forgot now what the leveldb vote index in ssb-db2 looked like in performance for the cold case, where it needs to build the leveldb before answering the query. Do you remember roughly what it was?

social db2 index on my machine is roughly 13 seconds.

staltz:
> social db2 index on my machine is roughly 13 seconds.

oh that's a lot, it's a big gain if we can take it down from 13 seconds cold to 2 seconds cold.

arj03:
Sure but that is for 3 indexes: vote, root and mentions. But yes, if we can get it down to 6 then 👍

arj03:
Base is actually the worst. Its like double of social.

staltz:
> Sure but that is for 3 indexes: vote, root and mentions. But yes, if we can get it down to 6 then +1

Oh right, it's one leveldb for all those. hmm...

staltz:
Idea: put a cache in `getRecord` and make also `getMessage` use getRecord's cache. Might be smart because `matchAgaintPrefix` anyway has to fetch the record to check if it's a false positive, and then again to deliver the results it does another fetch.

staltz:
@arj03 Check the last commit, I managed to shave off some milliseconds by avoiding Promises (they are based on the microscheduler which may delay things a tiny bit) and by avoiding extra for loops.

Now we're down to roughly **28ms**. Note that the image below also shows Profiler-specific functions:

Screenshot from 2020-12-03 13-55-28

index.js
staltz: So I think I'll put unlimited exponential jumps back, and for the storage solution, compression. What do u think?

index.js
staltz: Even better

index.js
arj03: There is a count in saveCoreIndex that could be used to just trim of the 0s at the end I think on save. As for compression maybe we can do that in a separate PR?

staltz:
ready for a final review!

staltz:
oh, one last task would be updating README

staltz:
updated readme

arj03:
I reviewed the code. Looks good. There is one idea I want to test before we merge and close this.

Superb work :rocket:

staltz:
I can do the tarr index file cut off of empty space in a separate PR

arj03:
Here is the thing that I wanted to test. Instead of looping over all the results to find all the values, sort the prefixes and then do a binary search on that. The idea is that when you load the vote index you can sort it once and then all queries for vote links will be faster. It seems that this + the optimization above means that we might even be faster than level :-)

Example code:

```
    const bitset = new TypedFastBitSet()
    const target = op.data.value
    const targetPrefix = target.readUInt32LE(0)
    const prefixIndex = indexes[op.data.indexName]

    console.time("sort")
    let t = {}
    for (let i = 0; i < prefixIndex.count; ++i) {
      let r = t[prefixIndex.tarr[i]] || []
      r.push(i)
      t[prefixIndex.tarr[i]] = r
    }

    let sortedKeys = Object.keys(t).sort((x, y) => x - y)
    console.timeEnd("sort")

    console.time("sort find")
    const start = bsb.ge(sortedKeys, targetPrefix)
    const end = bsb.gt(sortedKeys, targetPrefix)
    let res = sortedKeys.slice(start, end)
    res.forEach(x => bitset.add(x))
    console.log("results sorted", t[res])
    console.timeEnd("sort find")
  
    console.time("loopy")
    for (let i = 0; i < prefixIndex.count; ++i) {
      if (prefixIndex.tarr[i] === targetPrefix) bitset.add(i)
    }
    console.timeEnd("loopy")
    console.log("results loopy", bitset.array())
```
and output:

```
sort: 481.513ms
results sorted [ 308836, 504010, 611882, 916688 ]
sort find: 1.899ms
loopy: 13.774ms
results loopy [ 308836, 504010, 611882, 916688 ]
```

arj03:
Merging, we can do these kind of optimizations as separate PRs

staltz:
Very cool idea to sort it! can you make a release of this so I can test in db2? probably you will also test, but i want to test too because i want to have some fun

arj03:
Ha, yes but won't have time until later tonight. 

0.22 released. Will also add you to npm tonight so you can do release of jitdb and ssb-db2.

PR #53, opened 2020-12-03
cut off empty space when saving tarr file

staltz:
Reduces about 35% of the file size, which is some megabytes.

before

```
drwxrwxr-x 2 staltz staltz     4096 joulu  3 16:51  .
drwxr-xr-x 8 staltz staltz     4096 joulu  3 16:51  ..
-rw-rw-r-- 1 staltz staltz  8192008 joulu  3 16:51  offset.index
-rw-rw-r-- 1 staltz staltz  8192008 joulu  3 16:51  sequence.index
-rw-rw-r-- 1 staltz staltz 16384008 joulu  3 16:51  timestamp.index
```

after

```
drwxrwxr-x 2 staltz staltz     4096 joulu  3 16:52  .
drwxr-xr-x 8 staltz staltz     4096 joulu  3 16:52  ..
-rw-rw-r-- 1 staltz staltz  5330716 joulu  3 16:52  offset.index
-rw-rw-r-- 1 staltz staltz  5330716 joulu  3 16:52  sequence.index
-rw-rw-r-- 1 staltz staltz 10661424 joulu  3 16:52  timestamp.index
```

PR #55, opened 2020-12-03
When saving leave 10% empty space

arj03:
This is a continuation of #53. We don't want to have to immediate grow the log again after loading, especially not to double the size ;) This still saves significantly. 8mb to 6.3mb per file for me.


staltz (approved):

staltz:
Good one

PR #58, opened 2020-12-08
support record fields smaller than prefix length

staltz:
Working on `isRoot()` in db2 led me down a rabbit hole of bugs that we definitely need to fix, this is one of them.

Say you search for `msg.value.content.channel === "ui"` using a prefix index. The value `"ui"` is smaller than the 32 bit prefix, and this should be supported but it currently crashes. This PR pads some zeroes in that case.

Gladly I was using little endian anyway (if I had paid attention which endianness to pick some week(s) ago, I would have chosen big endian because it's more logical to me), because little endian makes it easy to pad zeroes **at the end** and this doesn't change the integer output.

The new test fails on the old implementation, passes with the new implementation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -218,6 +218,16 @@ module.exports = function (log, indexesPath) {
     else return false
   }
 
+  function safeReadUint32(buf) {
+    if (buf.length < 4) {
+      const bigger = Buffer.alloc(4)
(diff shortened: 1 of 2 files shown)


arj03:
Nice one.

PR #59, opened 2020-12-08
Falsy targets

staltz:
Problem: `isRoot()` in db2 wasn't working, not even in the `master` branch, and that's because jitdb needed these fixes. And when moving `isRoot()` to prefix, we would need even more changes to jitdb.

It's about `op.data.value` being undefined, and having logic that checks that when doing the queries.

This PR does a couple of other stuff too, e.g. I noticed that there is a duplicate logic `Buffer.isBuffer(x) ? x : Buffer.from(x)`: (1) in operators, (2) in sanitizeOpData. I ended up removing sanitizeOpData completely, and putting all "sanitization" logic in the operators, including `indexName`. This made `index.js` a bit simpler to work with, we can trust that when we get the operation object, it is ready for use.

Finally, I wanted to add support for `op.data.value === null` too, because eventually someone will write a query using db2 that has null, and it would be good if this wouldn't crash jitdb. So checking for absence is now done with any falsy value.

### DIFF
--- a/index.js
+++ b/index.js
@@ -208,11 +208,11 @@ module.exports = function (log, indexesPath) {
   }
 
   function checkValue(opData, buffer) {
-    const seekKey = opData.seek(buffer)
-    if (opData.value === undefined) return seekKey === -1
+    const fieldStart = opData.seek(buffer)
+    if (!opData.value) return fieldStart === -1
     else if (
-      ~seekKey &&
-      bipf.compareString(buffer, seekKey, opData.value) === 0
+      ~fieldStart &&
+      bipf.compareString(buffer, fieldStart, opData.value) === 0
     )
       return true
     else return false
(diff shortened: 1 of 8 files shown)


index.js
```diff
@@ -513,10 +498,16 @@ module.exports = function (log, indexesPath) {
     done((err, recs) => {
       // FIXME: handle error better, this cb() should support 2 args
       if (err) return console.error(err)
+      const seek = op.data.seek
```
staltz, line 501: This is tiny optimization somewhat unrelated to this PR, but good to have. Since the for-loop below is a hot path, doing many times `.` into object fields is wasteful, so we hold a reference to the seek function once.

arj03:
Yeah, makes the tests a bit more verbose, but I guess since we mostly using the convenience wrappers around jitdb it should be fine with the sanitize changes.

Great with the null check as well 👍

staltz:
Yeah, I thought about replacing the raw op objects with calls to the operators instead. That would be anyway reasonable to do, but I didn't want to change the tests too much. Would be good to do, but we got a bunch of other work to do

PR #60, opened 2020-12-08
simple benchmark scripts

staltz:
I did this one quickly before and during our Jitsi meeting. I want to close PR #21 but still salvage the good parts. This PR adds a new script, not run in CI, and not downloading the fixture (re-generating it instead).

looks like this:

```

> jitdb@0.23.0 benchmark /home/staltz/oss/manyverse/jitdb
> node benchmark/index.js | tap-spec


  core indexes

    ✔ duration: 8ms

  query one huge index (first run)

    ✔ duration: 631ms

  query one huge index (second run)

    ✔ duration: 257ms

  query three indexes (first run)

    ✔ duration: 418ms

  query three indexes (second run)

    ✔ duration: 202ms


  total:     5
  passing:   5
  duration:  1.6s
```

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -1,0 +1,145 @@
+const test = require('tape')
+const fs = require('fs')
+const path = require('path')
+const FlumeLog = require('async-flumelog')
+const generateFixture = require('ssb-fixtures')
+const rimraf = require('rimraf')
+const mkdirp = require('mkdirp')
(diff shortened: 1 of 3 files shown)


arj03:
Nice1

arj03:
Added a small noCreate flag you can use so it doesn't create the db on each run

PR #62, opened 2020-12-10
fix live() with empty ops

staltz:
- 1st commit adds the tests (and they fail as expected in CI)
- 2nd commit fixes the bug (and tests pass)

### DIFF
--- a/index.js
+++ b/index.js
@@ -658,6 +658,7 @@ module.exports = function (log, indexesPath) {
       else if (op.type === 'AND') ok = isValueOk(op.data, value, false)
       else if (op.type === 'OR') ok = isValueOk(op.data, value, true)
       else if (op.type === 'LIVEOFFSETS') ok = true
+      else if (!op.type) ok = true
 
       if (ok && isOr) return true
       else if (!ok && !isOr) return false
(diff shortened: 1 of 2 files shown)


test/operators.js
```diff
@@ -780,6 +780,34 @@ prepareAndRunTest('support empty deferred operations', dir, (t, db, raf) => {
   })
 })
 
+prepareAndRunTest('empty deferred AND equal', dir, (t, db, raf) => {
```
staltz, line 783: This test actually passes with the previous implementation, it was just good to add anyway.

index.js
```diff
@@ -810,7 +811,10 @@ module.exports = function (log, indexesPath) {
             })
           )
         } else {
-          const opts = { live: true, gt: seq }
+          const opts =
+            seq === -1
+              ? { live: true, gt: indexes['offset'].seq }
+              : { live: true, gt: seq }
```
staltz, line 817: if `seq === -1`, it means we didn't detect a seq, so instead of doing stream over `seq > -1` (which gives old values, and that might make some sense, except see toPullStream operator, it already runs jitdb.paginate for the old values, and after that, jitdb.live, so we really don't need old values)

arj03: Ah yes

Issue #33, opened 2020-11-18
Do something about timestamp


arj03:
There is the following FIXME in the code.
 
     // FIXME: maybe min of the two timestamps:
      // https://github.com/ssbc/ssb-backlinks/blob/7a731d03acebcbb84b5fee5f0dcc4f6fef3b8035/emit-links.js#L55

arj03:
Fixed in #66

PR #66, opened 2020-12-13
timestamp = min(declared, arrived)

staltz:
For issue #33 

- First commit adds the test, which fails for the previous implementation (see CI)
- Second commit adds the fix, which makes tests pass (see CI)


arj03:
Excellent

Issue #41, opened 2020-11-25
Possibly split LIVEOFFSETS so that it only supports passing a pull-stream


staltz:
This idea: https://github.com/ssb-ngi-pointer/ssb-db2/pull/37/files#r530435177

PR #68, opened 2020-12-14
split LIVEOFFSETS and OFFSETS

staltz:
For issue #41 

Basically they should be used with an OR now: `or(offsets([1,2,3], liveOffsets(ps))`. I can update ssb-db2 after this is merged.


arj03:
Looks good

Issue #65, opened 2020-12-13
Store index version


staltz:
If we change the internal logic for an index, then we should rebuild the index. E.g. the changes to timestamp index logic are just one example. We need to store in the index file a number that reflects the version. This should be done before putting it into production.

PR #69, opened 2020-12-14
add bitset.size() and version to index file header

staltz:
For issues #8 and #65, this adds two new fields to the file header: `version` and `bitsetSize`. Note it's a breaking change for any system currently deploying jitdb indexes.

So far there is no `bitset.size()` optimization (that old idea of not storing indexes that are too small), but if we ever want to implement that idea, this new header would make it easier.

### DIFF
--- a/files.js
+++ b/files.js
@@ -4,17 +4,38 @@ const TypedFastBitSet = require('typedfastbitset')
 const { readFile, writeFile } = require('atomically-universal')
 const toBuffer = require('typedarray-to-buffer')
 
-function saveTypedArrayFile(filename, seq, count, tarr, cb) {
+const FIELD_SIZE = 4 // bytes
+
+/*
+ * ## File format for tarr files
+ *
+ * Each header field is 4 bytes in size.
+ *
+ * | offset (bytes) | name    | type     |
+ * |----------------|---------|----------|
+ * | 0              | version | UInt32LE |
+ * | 4              | seq     | UInt32LE |
+ * | 8              | count   | UInt32LE |
+ * | 12             | (N/A)   | (N/A)    |
+ * | 16             | body    | Buffer   |
+ *
+ * Note that the 4th header field (offset=12) is empty on purpose, to support
+ * `new Float64Array(b,start,len)` where `start` must be a multiple of 8
+ * when loading in `loadTypedArrayFile`.
+ */
+
+function saveTypedArrayFile(filename, version, seq, count, tarr, cb) {
   if (!cb) cb = () => {}
 
   const dataBuffer = toBuffer(tarr)
   // we try to save an extra 10% so we don't have to immediately grow
   // after loading and adding again
   const saveSize = Math.min(count * 1.1, tarr.length)
-  const b = Buffer.alloc(8 + saveSize * tarr.BYTES_PER_ELEMENT)
-  b.writeInt32LE(seq, 0)
-  b.writeInt32LE(count, 4)
-  dataBuffer.copy(b, 8)
+  const b = Buffer.alloc(4 * FIELD_SIZE + saveSize * tarr.BYTES_PER_ELEMENT)
+  b.writeUInt32LE(version, 0)
+  b.writeUInt32LE(seq, FIELD_SIZE)
+  b.writeUInt32LE(count, 2 * FIELD_SIZE)
+  dataBuffer.copy(b, 4 * FIELD_SIZE)
 
   writeFile(filename, b)
     .then(() => cb())
@@ -24,11 +45,13 @@ function saveTypedArrayFile(filename, seq, count, tarr, cb) {
 function loadTypedArrayFile(filename, Type, cb) {
   readFile(filename)
     .then((buf) => {
-      const seq = buf.readInt32LE(0)
-      const count = buf.readInt32LE(4)
-      const body = buf.slice(8)
+      const version = buf.readUInt32LE(0)
+      const seq = buf.readUInt32LE(FIELD_SIZE)
+      const count = buf.readUInt32LE(2 * FIELD_SIZE)
+      const body = buf.slice(4 * FIELD_SIZE)
 
       cb(null, {
+        version,
         seq,
         count,
         tarr: new Type(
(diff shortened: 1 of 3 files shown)


files.js
```diff
@@ -41,21 +57,33 @@ function loadTypedArrayFile(filename, Type, cb) {
...
-    if (err) cb(err)
-    else {
-      const bitset = new TypedFastBitSet()
-      bitset.words = tarr
-      bitset.count = count
-      cb(null, { seq, bitset })
+  loadTypedArrayFile(
+    filename,
+    Uint32Array,
+    (err, { version, seq, count, tarr }) => {
```
arj03, line 77: you are missing bitsetSize here ;)

index.js
```diff
@@ -460,18 +475,20 @@ module.exports = function (log, indexesPath) {
...
-          index.tarr = tarr
           index.count = count
+          index.tarr = tarr
           index.lazy = false
           cb()
         }
       )
     } else {
-      loadBitsetFile(index.filepath, (err, { seq, bitset }) => {
+      loadBitsetFile(index.filepath, (err, { version, seq, bitset }) => {
```
arj03, line 489: and count should be here as well

arj03:
Had a crash when running example with this PR

staltz:
What was the crash?

arj03:
Somewhere in slowEqual. There is something there if you specify a path it can't find I think. This is unrelated to this PR, but something that would be nice to fix.

files.js
staltz: It's not necessary, right? bitsetSize is only relevant in `loadBitsetFile`, and actually it would only be necessary once we start using that field, for that idea #8

index.js
staltz: What's the reason? `count` was already available before this PR, and `count` was not used in this line of code

arj03:
To reproduce the problem comment out the first example in example.js, the one with the ship. Then make sure indexes are deleted and run node example. Then second run will fail with:

```
internal/validators.js:92
        throw new ERR_OUT_OF_RANGE(name, 'an integer', value);
        ^

RangeError [ERR_OUT_OF_RANGE]: The value of "sourceStart" is out of range. It must be an integer. Received NaN
    at Buffer.compare (buffer.js:877:5)
    at Object.compareString (/home/arj/dev/jitdb/node_modules/bipf/index.js:326:17)
    at checkEqual (/home/arj/dev/jitdb/index.js:246:12)
    at updateIndexValue (/home/arj/dev/jitdb/index.js:302:32)
    at Object.write (/home/arj/dev/jitdb/index.js:362:16)
    at Stream._writeToSink (/home/arj/dev/jitdb/node_modules/async-flumelog/stream.js:57:32)
    at Stream._handleBlock (/home/arj/dev/jitdb/node_modules/async-flumelog/stream.js:91:12)
    at /home/arj/dev/jitdb/node_modules/async-flumelog/stream.js:133:14
    at Request._callback (/home/arj/dev/jitdb/node_modules/async-flumelog/index.js:101:9)
    at Request.callback (/home/arj/dev/jitdb/node_modules/random-access-storage/index.js:173:8) {
  code: 'ERR_OUT_OF_RANGE'
}
```

staltz:
Updated this PR to not save bitsetSize, but note that this PR now **fails** tests.

There's one big change I made, which was the header fields `version`, `seq`, and `count` are now all `UInt32LE`. They used to be `Int32LE`. The tests are failing because some indexes are being saved with `seq=-1`. On the other hand, we want to **not** empty indexes! See #71.

So that's what I'm gonna do in this PR, tackle also #71 in a new commit soon. The reason why I changed to unsigned 32: we don't need negative numbers for these fields, and we gain a lot more range with unsigned. It seems we will need that larger range especially for `seq` (which is "size of the log file this index corresponds to"), and `Int32LE` would support only log files up to 2 GB big.

staltz:
Ready for review

files.js
```diff
@@ -4,17 +4,38 @@ const TypedFastBitSet = require('typedfastbitset')
...
 const toBuffer = require('typedarray-to-buffer')
 
-function saveTypedArrayFile(filename, seq, count, tarr, cb) {
+const FIELD_SIZE = 4 // bytes
+
+/*
+ * ## File format for tarr files
+ *
+ * Each header field is 4 bytes in size.
+ *
```
arj03, line 13: nice

arj03:
I tried this branch using the recipe [here](https://github.com/ssb-ngi-pointer/jitdb/pull/69#issuecomment-745107067) and still have the same problem. Its really wierd, the diff looks good. Have to double check it

arj03:
Ah, version and seq is switched for some reason

staltz:
I'll take a look at that problem, I'll try to fix it in this PR

arj03:
In save it seems:

```
jitdb saving core index: offset +1ms
saving version 1
saving seq 983320378
  jitdb saving core index: timestamp +2ms
saving version 1
saving seq 983320378
  jitdb saving core index: sequence +2ms
saving version 1
saving seq 983320378
  jitdb saving index: value_content_type_vote +3ms
saving version 983320378
saving seq 1
  jitdb saving index: value_content_vote_expression_u26F5 +2ms
saving version 983320378
saving seq 1
  jitdb saving index: value_author_@+UMKhpbzXAII+2x2F7ZlsgkJwIsxdfeFi36Z5Rk1gCfY0=x2Eed25519 +2ms
saving version 983320378
saving seq 1
```

files.js
```diff
@@ -41,21 +64,25 @@ function loadTypedArrayFile(filename, Type, cb) {
     .catch(cb)
 }
 
-function saveBitsetFile(filename, seq, bitset, cb) {
+function saveBitsetFile(filename, version, seq, bitset, cb) {
   bitset.trim()
-  saveTypedArrayFile(filename, seq, bitset.count, bitset.words, cb)
+  saveTypedArrayFile(filename, seq, version, bitset.count, bitset.words, cb)
```
arj03, line 69: here is the problem :)

staltz: aaaah

staltz: weird that tests don't break, I'll try adding a failing test first

staltz:
Added test that fails because of swapped version and seq.

arj03:
Working now :)

PR #70, opened 2020-12-14
Benchmark paginate, enable CI

staltz:


arj03:
Trying to understand what is going on here, its a bit new to me :) if I understand this correctly the actions bot with post this kind of message when it has finished running a CI test?

staltz:
There's a new CI job called `benchmark` that runs after the 3 test jobs complete (successfully). The `benchmark` job does `npm run benchmark` and then after that posts a comment here with the results of that run

arj03:
Great :-)

PR #72, opened 2020-12-14
add test for live() bug

staltz:
1st commit adds failing test, 2nd commit adds bug fix.

### DIFF
--- a/test/live-then-grow.js
+++ b/test/live-then-grow.js
@@ -1,14 +1,10 @@
-const test = require('tape')
-const fs = require('fs')
 const validate = require('ssb-validate')
 const ssbKeys = require('ssb-keys')
 const path = require('path')
-const FlumeLog = require('async-flumelog')
 const rimraf = require('rimraf')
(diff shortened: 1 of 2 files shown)


test/live-then-grow.js
```diff
@@ -0,0 +1,61 @@
...
+  slowEqual,
+} = require('../operators')
+
+const dir = '/tmp/jitdb-live-then-grow'
+rimraf.sync(dir)
+mkdirp.sync(dir)
+
+var keys = ssbKeys.loadOrCreateSync(path.join(dir, 'secret'))
+
+test('live from empty log', (t) => {
```
staltz, line 27: I can update this to use prepareAndRun, I actually didn't use it here because I thought I would need to close jitdb and reopen it, but it turned out I didn't need to do that.

but i'll do this update tomorrow, unless you want to do it

arj03: Ok, this looks good. I'll wait for the update and merge

arj03:
Sweet, nice find!

PR #74, opened 2020-12-15
Better support for delete

arj03:
For delete in ssb-db2 we need a way to get the seq of specific messages so we can call delete on the log with the seq. While testing this in ssb-db2 I found some other bugs, one in matchprefix and one related to https://github.com/flumedb/async-flumelog/pull/13. Because of this the PR is bigger than I would have liked. I didn't add an operator for this because the query is simple enough to just write using the raw api and can always be added later if needed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -396,7 +396,7 @@ module.exports = function (log, indexesPath) {
         }
     })
 
-    let offset = -1
+    let offset = 0
 
     let updatedOffsetIndex = false
     let updatedTimestampIndex = false
@@ -406,14 +406,16 @@ module.exports = function (log, indexesPath) {
     log.stream({}).pipe({
       paused: false,
       write: function (record) {
-        offset++
-
         const seq = record.seq
         const buffer = record.value
 
         if (updateOffsetIndex(offset, seq)) updatedOffsetIndex = true
 
-        if (!buffer) return // deleted
+        if (!buffer) {
+          // deleted
+          offset++
+          return
+        }
 
         if (updateTimestampIndex(offset, seq, buffer))
           updatedTimestampIndex = true
@@ -435,6 +437,8 @@ module.exports = function (log, indexesPath) {
           else
             updateIndexValue(op, newIndexes[op.data.indexName], buffer, offset)
         })
+
+        offset++
       },
       end: () => {
         const count = offset // incremented at end
--- a/index.js
+++ b/index.js
@@ -335,25 +335,22 @@ module.exports = function (log, indexesPath) {
       write: function (record) {
         if (updateOffsetIndex(offset, record.seq)) updatedOffsetIndex = true
 
-        if (record.value != null) {
-          // not deleted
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -335,16 +335,25 @@ module.exports = function (log, indexesPath) {
...
-          updatedTimestampIndex = true
-
-        if (updateSequenceIndex(offset, record.seq, record.value))
-          updatedSequenceIndex = true
-
-        if (indexNeedsUpdate) {
-          if (op.data.prefix)
-            updatePrefixIndex(op.data, index, record.value, offset, record.seq)
-          else updateIndexValue(op, index, record.value, offset)
+        if (record.value != null) {
```
staltz, line 338: How about `if (!record.value) return // deleted record`? It reduces the amount of nested ifs, which can get hard to read

index.js
```diff
@@ -563,6 +580,11 @@ module.exports = function (log, indexesPath) {
       const seek = op.data.seek
       for (let i = 0, len = recs.length; i < len; ++i) {
         const { value, offset } = recs[i]
+        if (value === undefined) {
```
staltz, line 583: I prefer `if (!value)` because it covers both `null` and `undefined` in case a null ever sneaks in. Or if you want to avoid all other falsy values, `if (value == null)` covers only `null` and `undefined`

index.js
arj03: we still need the offset++, but I guess I could do that in the return

staltz: Yeah, I thought that maybe the initial value of offset could be `-1`, and then first thing inside the stream iteration we do is `offset++`. That's one way, I'm not saying it has to be that way

index.js
```diff
@@ -390,7 +396,7 @@ module.exports = function (log, indexesPath) {
         }
     })
 
-    let offset = 0
+    let offset = -1
```
staltz, line 399: I'm so sorry for going back and forth with this one, but maybe it would be better after all if it started with 0, and using the same pattern as in `updateIndex`, because there we needed to find the next possible offset (it may not be 0).

So perhaps `offset = 0`, and then using the same `if(!buffer) { offset++; return }` trick?

The concern I have is just consistency, that we use roughly the same patterns in `updateIndex` and `createIndexes`, to make it easier to read code, as well as avoid off by one errors with this -1 initialization.

staltz (changes requested):
Just one more thing and then I'm happy to merge. I'm actually glad that adding support for delete is rather straightforward, just a couple of `if`s here and there. And great to have tests, I think our tests are a very valuable asset as we go forwards.

index.js
arj03: Alright :-)

staltz (approved):
This PR looks all good now, and I think it could be even put out of draft and merged, because I already released jitdb 0.30.0, so this merged one could go into 0.31.0

cblgh:
sick idea of using github actions for running automated benchmarks

staltz (approved):
:ship: it

PR #76, opened 2020-12-16
Change live() operator to mean live only

### LINKED ISSUE #75: operator `liveOnly()`
Very similar to `live()`, but sets `meta.liveOnly = true`, and in toPullStream() we check meta.live and meta.liveOnly, and do the pull-cat accordingly.

Use case: when we don't want old values in the live stream.

@arj03 Could you do this one? (PS: as I'm putting ssb-db2 in manyverse i'm mapping out what's still needed)

arj03:
And support old values using {old:true} opts. This should fix #75

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,15 +352,33 @@ const {
 
 ## Prefix indexes
 
-Most indexes in JITDB are bitvectors, which are suitable for answering boolean queries such as "is this msg a post?" or "is this msg from author A?". For each of these queries, JITDB creates one file.
+Most indexes in JITDB are bitvectors, which are suitable for answering
+boolean queries such as "is this msg a post?" or "is this msg from
+author A?". For each of these queries, JITDB creates one file.
 
-This is fine for several cases, but some queries are not boolean. Queries on bitvectors such as "is this msg a reply to msg X?" can end up generating `N` files if the "msg X" can have N different values. The creation of indexes is this case becomes the overhead.
+This is fine for several cases, but some queries are not
+boolean. Queries on bitvectors such as "is this msg a reply to msg X?"
+can end up generating `N` files if the "msg X" can have N different
+values. The creation of indexes is this case becomes the overhead.
 
-**Prefix indexes** help in that case because they can answer non-boolean queries with multiple different values but using just one index file. For example, for N different values of "msg X", just one prefix index is enough for answering "is this msg a reply to msg X?".
+**Prefix indexes** help in that case because they can answer
+non-boolean queries with multiple different values but using just one
+index file. For example, for N different values of "msg X", just one
+prefix index is enough for answering "is this msg a reply to msg X?".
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -352,15 +352,33 @@ const {
 
 ## Prefix indexes
 
-Most indexes in JITDB are bitvectors, which are suitable for answering boolean queries such as "is this msg a post?" or "is this msg from author A?". For each of these queries, JITDB creates one file.
```
arj03, line 355: Couldn't resist, changed nothing here, just meta-q

test/operators.js
```diff
@@ -827,12 +826,11 @@ prepareAndRunTest('support live offset operations', dir, (t, db, raf) => {
...
     ),
     toPullStream(),
-    pull.filter((x) => x), // filter out the first non-live empty result
     pull.drain((msg) => {
-      t.equal(msg.value.author, bob.id)
+      t.equal(msg.value.author, alice.id)
 
       // test we don't get live messages after aborting stream
-      addMsg(state.queue[1].value, raf, (e2, msg2) => {
+      addMsg(state.queue[0].value, raf, (e2, msg2) => {
```
staltz, line 833: What's going on here, why are we adding `state.queue[0].value` twice, and where did bob go?

operators.js
```diff
@@ -396,9 +397,14 @@ function toPullStream() {
...
-      pull.map((ops) =>
-        cat([paginateStream(ops), meta.live ? meta.db.live(ops) : null])
-      ),
+      pull.map((ops) => {
+        if (meta.liveOnly) return meta.db.live(ops)
+        else
+          return cat([
+            paginateStream(ops),
+            meta.live ? meta.db.live(ops) : null,
+          ])
```
staltz, line 406: This is totally fine implementation, but to nitpick a tiny bit, how about an `else if (meta.live) return cat([.....])` and then an `else return paginateStream(ops)`? Just to satisfy the eye :)

staltz:
Good! The `live({ old: true })` is also my preferred solution. :)

operators.js
```diff
@@ -283,8 +283,9 @@ function or(...args) {
   }
 }
 
-function live() {
-  return (ops) => updateMeta(ops, 'live', true)
+function live(opts) {
+  if (opts && opts.old) return (ops) => updateMeta(ops, 'live', true)
```
staltz, line 287: Actually, because this `meta.live` is only internal, it would be good to rename these:

- live => liveAndOld
- liveOnly => live (because this is the default case `live()` and the semantics of the default case is "live only")

test/operators.js
arj03: It should do the same thing as before, we just don't need to skip the values. Before it was also adding the same message 2 times. I'll add back bob because that should make the test easier to read.

operators.js
arj03: Good idea

test/operators.js
staltz: Oh ok I see. Thanks

arj03:
Fixed, that was a good review. Code feels cleaner now.

staltz:
Good to merge, then! Nice that you added more tests too :)

arj03:
Moar!

staltz:
By the way, this looks like a breaking change and we may need to update ssb-db2 too. You might know it, just saying because we need a written down reminder somewhere.

PR #77, opened 2020-12-17
test that live() overrides previous live()

staltz:
This one actually was drafted in my dreams at night.

First commit adds failing test, second commit adds fix that passes the test.

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -284,8 +284,8 @@ function or(...args) {
 }
 
 function live(opts) {
-  if (opts && opts.old) return (ops) => updateMeta(ops, 'liveAndOld', true)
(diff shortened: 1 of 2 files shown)


arj03:
Alrighty, not sure if anyone would do that, but no harm in getting this check in :)

staltz:
@arj03 They might not do it like the tests show, but maybe they could do it like this and then it's more sneaky:

```js
const collection = db.query(
  and(type('post'), author(me)),
  live()
)

pull(
  db.query(
    collection,
    live({ old: true }),
    toPullStream()
  ),
  pull.drain(console.log)
)
```

PR #79, opened 2020-12-25
Use async-log

arj03:
Needs https://github.com/ssb-ngi-pointer/async-log/pull/14


staltz (approved):
Nice to delete those comments :)

PR #80, opened 2020-12-30
Fix error handling when loading files

arj03:
Still trying to make things work in the browser. I'm running into all kinds of nice error cases here ;-)

If file loading errors, don't try to destructure.

I also sneaked in a updated toBuffer as that is the version we are using in ssb-db2. It's mostly dropping IE support, which we don't need anyway in the browser. Shouldn't have any diff in node.


staltz (approved):
Looks good!

arj03:
Thanks

At least merging feels productive ;-)

PR #82, opened 2021-01-13
Fix seq index not in sync with other indexes

arj03:
This should fix https://github.com/ssb-ngi-pointer/async-append-only-log/issues/17. I tested in with the rm thing on example and can confirm that this fixes the problem.


staltz (approved):
Great, I hope this makes a different in prod :)

arj03:
It should :)

PR #87, opened 2021-01-16
Fix an off by 1 in when to grow

arj03:
I was testing adding reactions and noticed that something was really off. My reactions was not showing up properly when reloading. Turns out there was an off by 1 in the grow code that meant the first entry written would be 0 because you would write into the nothingness and the next value would be written ok, but the space where the first value was supposed to be was not a 0.

### DIFF
--- a/test/prefix.js
+++ b/test/prefix.js
@@ -5,6 +5,8 @@ const { prepareAndRunTest, addMsg, helpers } = require('./common')()
 const rimraf = require('rimraf')
 const mkdirp = require('mkdirp')
 
+const jitdb = require('../index')
+
 const dir = '/tmp/jitdb-prefix'
 rimraf.sync(dir)
 mkdirp.sync(dir)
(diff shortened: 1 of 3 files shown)


test/prefix.js
```diff
@@ -111,3 +113,86 @@ prepareAndRunTest('Prefix equal falsy', dir, (t, db, raf) => {
...
+  }
+
+  addMsg(state.queue[0].value, raf, (err, msg) => {
+    addMsg(state.queue[1].value, raf, (err, msg) => {
+      addMsg(state.queue[2].value, raf, (err, msg) => {
+        db.all(voteQuery, 0, false, false, (err, results) => {
+          t.equal(results.length, 1)
+          t.equal(results[0].value.content.type, 'vote')
+
+          db = jitdb(raf, path.join(dir, 'indexes' + name))
```
staltz, line 173: What's the idea here with reinitializing the jitdb instance?

arj03: It was easier to trigger the bug this way because of the index trim. Also this is much closer to a realistic test than writing 32k values 😊

staltz (approved):

PR #89, opened 2021-01-18
test that we can change operators to prefix

staltz:
In preparing to change `authors()` to prefix ( https://github.com/ssb-ngi-pointer/ssb-db2/issues/123 ), I wanted to add this test and fix the implementation so that there are no side-effects (e.g. maybe some crash by trying to load a typedfastbitset as if it were a prefix index). Gladly, when I wrote the test, the current implementation already passed (because of `op.data.indexName`), so no fix needed.

### DIFF
--- a/test/prefix.js
+++ b/test/prefix.js
@@ -47,6 +47,56 @@ prepareAndRunTest('Prefix equal', dir, (t, db, raf) => {
   })
 })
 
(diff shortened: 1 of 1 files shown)


arj03 (approved):

arj03:
Perfect

PR #90, opened 2021-01-18
add test for skipping deleted records in prefix indexes

staltz:
1st commit adds failing test, 2nd commit fix

### DIFF
--- a/index.js
+++ b/index.js
@@ -566,8 +566,8 @@ module.exports = function (log, indexesPath) {
     opOffsets.sort((x, y) => x - y)
     const opOffsetsLen = opOffsets.length
     const { tarr } = indexes['seq']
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -758,7 +758,7 @@ module.exports = function (log, indexesPath) {
   function getRecord(seq, cb) {
     const offset = indexes['seq'].tarr[seq]
     log.get(offset, (err, value) => {
-      if (err && err.code === 'flumelog:deleted') cb()
+      if (err && err.code === 'flumelog:deleted') cb(null, { seq, offset })
```
staltz, line 761: This is the only thing needed for the fix, the other changes are style changes.

arj03: nice find

PR #91, opened 2021-01-18
Update TypedFastBitset to 0.2.1

staltz:
They actually shipped a new version today that fixed the bug.

I didn't notice any performance difference when running the benchmark locally.

### DIFF
--- a/index.js
+++ b/index.js
@@ -556,7 +556,7 @@ module.exports = function (log, indexesPath) {
     ensureIndexSync({ data: { indexName: 'sequence' } }, () => {
       const bitset = new TypedFastBitSet()
       const { count } = indexes['sequence']
-      for (var i = 0; i < count; ++i) bitset.add(i)
+      bitset.addRange(0, count)
       cb(bitset)
     })
   }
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -556,7 +556,7 @@ module.exports = function (log, indexesPath) {
     ensureIndexSync({ data: { indexName: 'sequence' } }, () => {
       const bitset = new TypedFastBitSet()
       const { count } = indexes['sequence']
-      for (var i = 0; i < count; ++i) bitset.add(i)
+      bitset.addRange(0, count)
```
staltz, line 559: Nice new API they added

package.json
```diff
@@ -26,7 +26,7 @@
     "sanitize-filename": "^1.6.3",
     "traverse": "^0.6.6",
     "typedarray-to-buffer": "^4.0.0",
-    "typedfastbitset": "~0.1.5"
+    "typedfastbitset": "~0.2.1"
```
arj03, line 29: we should probably do ^ here again, will fix that after merge.

staltz: I actually prefer the tilde by default, since it's common to get breaking changes in the minor num, in the npm ecosystem. That was the case of typedfastbitset 0.2.0 too.

arj03: Oh, will revert then. Just pushed a minor change

PR #94, opened 2021-01-21
support and(live()) operations 

staltz:
This is the bulk of the work to fix https://github.com/ssb-ngi-pointer/ssb-db2/issues/133

Note that the 2nd commit is a refactor that just shuffles around stuff. It's probably easiest to just review the 1st commit: https://github.com/ssb-ngi-pointer/jitdb/commit/97e7bb3de27c4c3f077010d6621efbefbb4b2ff4

About the refactor commit, `//#region` is a thing that VSCode interprets and allows you to collapse/expand. For non-VSCode users, this may still be readable comments. For VSCode users (many are), it might be helpful to collapse/expand.

Screenshot from 2021-01-21 15-50-18

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -251,12 +251,12 @@ function extractMeta(orig) {
 }
 
 function and(...args) {
-  const rhs = args
-    .map((arg) => (typeof arg === 'function' ? arg() : arg))
-    .filter((arg) => !!arg)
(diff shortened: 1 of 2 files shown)


arj03:
Can confirm this works. Thank you so much :)

Issue #56, opened 2020-12-03
Inverted indexes


staltz:
I just had an idea that is either silly/naive or groundbreaking.

This was inspired by your sort idea #54. So, normal indexes in jitdb can be considered a hash map from offset to boolean:

```
0 -> 0
1 -> 0
2 -> 1
3 -> 0
4 -> 1
5 -> 1
6 -> 0
7 -> 1
```

But prefix indexes are:

```
0 -> 0
1 -> %c9A
2 -> 0
3 -> %bWw
4 -> %c9A
5 -> 0
6 -> %zy7
7 -> %bWw
```

And the sort idea is just a way of squishing all the similar values together. What we need to do instead is *invert* the hash map above, from prefixes to offsets:

```
%c9A -> [1, 4]
%bWw -> [3, 7]
%zy7 -> [6]
```

Now we don't even need to sort, we get all the values that match the msgid. And look at all that space eliminated by not storing the zeroes. Since we saved up so much space, we can actually quit the prefix and go back to full id:

```
%c9Ac8fc8f0Aqmcuz/8sc=.ed25519 -> [1, 4]
%bWw39c91CIelu1285riC=.ed25519 -> [3, 7]
%zy718ci0aZmc58aTFUC0=.ed25519 -> [6]
```

Which eliminates the need to do a second iteration to remove false positives!

Considerations: the normal index doesn't directly store the offsets because those are implicit in the arrangement of the values, but inverted index stores all the offsets plus the full key, so it "uses more space". On the other hand, we're not storing all those zeroes, and there sure was a lot of zeroes. Might be interesting to test this in practice though.

Now, the inverted index above really looks like a leveldb kind of thing, so maybe we're just reinventing leveldb. Or maybe not! We don't need to sort the keys, and leveldb has strict sorting of keys lexicographically, and sorting isn't free. We just need to load the whole thing into memory and we don't care about sorting. 

Thoughts @arj03 ?

staltz:
More about the storage of zeroes: for instance take the conventional index `value_author_andrestaltzphone`. I ran the statistics here, and it's 97% zeroes, and this is 166KB file. Now do that times N authors. Assume we have 16k authors here on my laptop, then you'd get in total 16k files totalling 2.6GB, of which 2.3GB is zeroes. I'm pretty sure that eliminating all those zeroes would and using an inverted index get us down to megabytes. If it's like a single 100MB file, we can chop that into pieces.

arj03:
[...]

PR #95, opened 2021-01-22
Optimize prefixes for queries

arj03:
This took a while to wrap my head around. Basically before we were making it easy to create prefix indexes (just insert into tarr), but slow to query them loop. Instead lets optimize for the query because prefix indexes are used for very often used queries (key, votes, hasRoot etc.). The natural choice was to store prefixes as maps instead of arrays. Mapping the lookup key to sequences. Besides a bit slower index creation (25-70%) the size has almost doubled. I did not find a better solution than just storing them JSON stringified, but I'm sure there is a better way because we are just storing numbers here.

Query numbers for already created indexes using [perf](https://github.com/ssb-ngi-pointer/ssb-db2/pull/136) script:

## old

```
running: key initial
query: 130.071ms
running: key 2
query: 16.11ms
running: key again
query: 14.645ms
running: latest root posts
query: 88.219ms
running: latest posts
query: 166.388ms
running: votes initial
query: 43.506ms
running: votes 2
query: 11.124ms
running: votes again
query: 4.939ms
running: author posts
query: 225.619ms
running: author posts again
query: 39.766ms
```

## new

```
running: key initial
query: 294.774ms
running: key 2
query: 10.437ms
running: key again
query: 8.192ms
running: latest root posts
query: 77.361ms
running: latest posts
query: 137.998ms
running: votes initial
query: 93.16ms
running: votes 2
query: 0.902ms
running: votes again
query: 0.796ms
running: author posts
query: 414.573ms
running: author posts again
query: 33.201ms
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -326,8 +326,6 @@ module.exports = function (log, indexesPath) {
       if (fieldStart) {
         const buf = bipf.slice(buffer, fieldStart)
         addToPrefixMap(index.map, seq, buf.length ? safeReadUint32(buf) : 0)
-      } else {
-        addToPrefixMap(index.map, seq, 0)
       }
 
       index.offset = offset
--- a/index.js
+++ b/index.js
@@ -11,6 +11,8 @@ const debug = require('debug')('jitdb')
 const {
   saveTypedArrayFile,
   loadTypedArrayFile,
+  savePrefixMapFile,
+  loadPrefixMapFile,
   saveBitsetFile,
   loadBitsetFile,
   safeFilename,
@@ -113,6 +115,17 @@ module.exports = function (log, indexesPath) {
               filepath: path.join(indexesPath, file),
             }
             cb()
+          } else if (file.endsWith('.32prefixmap')) {
+            // Don't load it yet, just tag it `lazy`
+            indexes[indexName] = {
+              offset: -1,
+              count: 0,
+              map: {},
+              lazy: true,
+              prefix: 32,
+              filepath: path.join(indexesPath, file),
+            }
+            cb()
           } else if (file.endsWith('.index')) {
             // Don't load it yet, just tag it `lazy`
             indexes[indexName] = {
@@ -177,6 +190,21 @@ module.exports = function (log, indexesPath) {
       prefixIndex.offset,
       count,
       prefixIndex.tarr,
+      cb
+    )
+  }
+
+  function savePrefixMapIndex(name, prefixIndex, count, cb) {
+    if (prefixIndex.offset < 0) return
+    debug('saving prefix map index: %s', name)
+    const num = prefixIndex.prefix
+    const filename = path.join(indexesPath, name + `.${num}prefixmap`)
+    savePrefixMapFile(
+      filename,
+      prefixIndex.version || 1,
+      prefixIndex.offset,
+      count,
+      prefixIndex.map,
       cb
     )
   }
@@ -284,6 +312,29 @@ module.exports = function (log, indexesPath) {
     }
   }
 
+  function addToPrefixMap(map, seq, value) {
+    if (value === 0) return
+
+    let arr = map[value] || []
+    arr.push(seq)
+    map[value] = arr
+  }
+
+  function updatePrefixMapIndex(opData, index, buffer, seq, offset) {
+    if (seq > index.count - 1) {
+      const fieldStart = opData.seek(buffer)
+      if (fieldStart) {
+        const buf = bipf.slice(buffer, fieldStart)
+        addToPrefixMap(index.map, seq, buf.length ? safeReadUint32(buf) : 0)
+      } else {
+        addToPrefixMap(index.map, seq, 0)
+      }
+
+      index.offset = offset
+      index.count = seq + 1
+    }
+  }
+
   function updatePrefixIndex(opData, index, buffer, seq, offset) {
     if (seq > index.count - 1) {
       if (seq > index.tarr.length - 1) growTarrIndex(index, Uint32Array)
@@ -368,7 +419,9 @@ module.exports = function (log, indexesPath) {
           updatedSequenceIndex = true
 
         if (indexNeedsUpdate) {
-          if (op.data.prefix)
+          if (op.data.prefix && op.data.useMap)
+            updatePrefixMapIndex(op.data, index, buffer, seq, offset)
+          else if (op.data.prefix)
             updatePrefixIndex(op.data, index, buffer, seq, offset)
           else updateIndexValue(op, index, buffer, seq)
         }
@@ -389,7 +442,10 @@ module.exports = function (log, indexesPath) {
 
         index.offset = indexes['seq'].offset
         if (indexNeedsUpdate) {
-          if (index.prefix) savePrefixIndex(op.data.indexName, index, count)
+          if (index.prefix && index.map)
+            savePrefixMapIndex(op.data.indexName, index, count)
+          else if (index.prefix)
+            savePrefixIndex(op.data.indexName, index, count)
           else saveIndex(op.data.indexName, index)
         }
 
@@ -401,7 +457,14 @@ module.exports = function (log, indexesPath) {
   function createIndexes(opsMissingIndexes, cb) {
     const newIndexes = {}
     opsMissingIndexes.forEach((op) => {
-      if (op.data.prefix)
+      if (op.data.prefix && op.data.useMap) {
+        newIndexes[op.data.indexName] = {
+          offset: 0,
+          count: 0,
+          map: {},
+          prefix: typeof op.data.prefix === 'number' ? op.data.prefix : 32,
+        }
+      } else if (op.data.prefix)
         newIndexes[op.data.indexName] = {
           offset: 0,
           count: 0,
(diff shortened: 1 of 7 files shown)


staltz:
Oh, as a Map, it's basically an [inverted index](https://github.com/ssb-ngi-pointer/jitdb/issues/56) now! I'll review and think about this.

arj03:
After my talk with @staltz I did some tests where I tried to create bitsets on top of the prefix arrays and using the bitsets to narrow down the number of rows we need to run through. Note that modern CPUs are blazingly fast at running through arrays so we are up against 10ms for 1.5million.

The bitsets for one example query looks like this:

```
1434795 81918
0 8
1434795 81918
0 8
0 8
1434795 81918
0 8
0 8
739229 81918
717741 81918
649757 81918
581403 81918
716629 81918
852737 81918
1166050 81918
0 8
739782 81918
716175 81918
650761 81918
582353 81918
718284 81918
852146 81918
1165671 81918
0 8
740033 81918
717083 81918
650634 81918
582317 81918
718617 81918
851702 81918
1166308 81918
```

the first number being the number of number of matches, the second the size of the bitset in bytes.

Using the 8 "best" bitsets we get the following results:

```
found 25
found 27
found 28
found 30
union bitsets: 11.472ms
target prefix union 45058
loopy: 7.105ms
```

Meaning it used 4 of those bitsets to narrow down the result to 45k instead of 1.5m. Sadly the union operation is about the same speed as just running over the array, so I'm not sure if that idea is worth persuing more.

I'll fix whatever is making the benchmark tests not run and see if we can save this to storage in a better way. There is the possibility of doing this prefix map as another type, so we have the existing prefix also.

arj03:
Another idea we discussed was not saving the 0 (key) buckets for the map implementation. Going to test that.

staltz:
I actually worked on this branch for 10mins just to put the perf.js inside benchmark/index.js, and it worked. I didn't commit and push because I had to go out, but the code is ready. I'll later rebase onto this branch, so you don't need to do it. :) @arj03

arj03:
Sweet. Thank you :)

staltz:
Aaaactually that was on db2, not in jitdb. But still, probably useful.

arj03:
So turns out the skip 0 is a really good idea for things where not all messages have values and you don't need to query for unknown. So like votes.

Before:

```
-rw-rw-r-- 1 arj arj 14233563 Jan 22 15:18 ssb/db2/indexes/key.32prefixmap
-rw-rw-r-- 1 arj arj 10661035 Jan 22 15:18 ssb/db2/indexes/value_author.32prefixmap
-rw-rw-r-- 1 arj arj 12015680 Jan 22 15:18 ssb/db2/indexes/value_content_vote_link.32prefixmap
```

After:

```
-rw-rw-r-- 1 arj arj 14233563 Jan 22 15:17 ssb/db2/indexes/key.32prefixmap
-rw-rw-r-- 1 arj arj 10661035 Jan 22 15:17 ssb/db2/indexes/value_author.32prefixmap
-rw-rw-r-- 1 arj arj  4634479 Jan 22 15:17 ssb/db2/indexes/value_content_vote_link.32prefixmap
```

arj03:
That translates of course into a bit faster initial load on votes if used. I'm not exactly sure where to go next with this. For votes this really makes a lot of sense. The indexes are smaller, and much faster. Not really for author so maybe add a new index type?

Existing prefix:

```
6,4M	ssb/db2/indexes/value_content_vote_link.32prefix
```

staltz:
Yep, I'm all for making different types of indexes and tailoring the operators to use different types of indexes.

arj03:
@staltz this should now be ready. I ended up with a new useMap flag that will then create a new file. I tested that going from a ordinary prefix index to a map prefix index works. This should be good for `votesFor` and `hasRoot`. I was also thinking that this could be really good for [about](https://github.com/ssbc/ssb-social-index/pull/4) and for [friends lite](https://github.com/ssb-ngi-pointer/ssb-db2/issues/5).

staltz:
Nice! I checked the code and looks neat so far. I have a few more things I'd like to review.

Other question: have you tried BIPF instead of JSON.parse/stringify?

arj03:
Right I forgot to say that. I did. Sadly its not faster, especially for the saving part and we don't get any of the "I can just get this part", we really need to decode the whole thing.

staltz:
@arj03 I added a commit to benchmark prefix map indexes in CI. If I understood correctly, the 1st run of a prefix map is slightly slower than the 1st run of a prefix non-map index, but the 2nd run is noticeably faster (from ~16ms down to ~4ms). Is that your experience as well?

(CI will report bench results soon)

arj03:
Exactly

index.js
```diff
@@ -284,6 +312,29 @@ module.exports = function (log, indexesPath) {
...
+  }
+
+  function updatePrefixMapIndex(opData, index, buffer, seq, offset) {
+    if (seq > index.count - 1) {
+      const fieldStart = opData.seek(buffer)
+      if (fieldStart) {
+        const buf = bipf.slice(buffer, fieldStart)
+        addToPrefixMap(index.map, seq, buf.length ? safeReadUint32(buf) : 0)
+      } else {
+        addToPrefixMap(index.map, seq, 0)
```
staltz, line 330: I wonder what's the purpose of doing `map = {0: [seqA, seqB, seqC, seqD], ...}` when the `fieldStart` is zero? `fieldStart === 0` should mean that the field is at the very beginning of the `buffer`? @arj03

staltz: Oh, and then `addToPrefixMap(index.map, seq, 0)` would actually *not* do anything because of line 315. I think this is somewhat confusing. Maybe we should do

```js
if (~fieldStart) {

}
```
with no `else`

arj03: You mean the else block? Yeah, that's a bit wasteful, could just be removed. addToPrefixMap has a guard against 0 because that saves a ton of space in these kinds of indexes.

arj03: Fixed

index.js
```diff
@@ -284,6 +312,27 @@ module.exports = function (log, indexesPath) {
...
+
+    let arr = map[value] || []
+    arr.push(seq)
+    map[value] = arr
+  }
+
+  function updatePrefixMapIndex(opData, index, buffer, seq, offset) {
+    if (seq > index.count - 1) {
+      const fieldStart = opData.seek(buffer)
+      if (fieldStart) {
```
staltz, line 326: @arj03 Thanks. Did you also see my comment about `~fieldStart`? Or is this intentionally testing for "field must not be at the beginning of the buffer"?

staltz:
Note, `slowEqual` also needs to support `useMap`

index.js
arj03: It was keeping it in line with normal prefix indexes. It seems you are correct in that both have a bug if its at position 0 in the buffer.

operators.js
```diff
@@ -107,16 +107,21 @@ function equal(seek, target, opts) {
...
   const valueName = !value ? '' : value.toString()
   const indexType = opts.indexType
-  const indexName = opts.prefix
-    ? safeFilename(indexType)
-    : safeFilename(indexType + '_' + valueName)
+  const indexName = safeFilename(
+    opts.prefix
+      ? opts.useMap
+        ? indexType + '_map'
+        : indexType
```
staltz, line 114: I think this part is where my main review is. Suppose we have a query like `slowEqual('value.content.geolocation.map', 'mercator', { prefix: 32 })`. But assuming that there are no schema enforcement in SSB, there would also be `slowEqual('value.content.geolocation', 'mercator', { prefix: 32 })`

Then prior to this PR, the `indexName` for those would be:

- For ``slowEqual('value.content.geolocation.map', 'mercator', { prefix: 32 })`
  - value_content_geolocation_map`
- For `slowEqual('value.content.geolocation', 'mercator', { prefix: 32 })`
  - `value_content_geolocation`

But after this PR, if we change

```diff
-slowEqual('value.content.geolocation', 'mercator', { prefix: 32 })
+slowEqual('value.content.geolocation', 'mercator', { prefix: 32, useMap: true })
```

Then the `indexName`s would be 

- For ``slowEqual('value.content.geolocation.map', 'mercator', { prefix: 32 })`
  - `value_content_geolocation_map`
- For `slowEqual('value.content.geolocation', 'mercator', { prefix: 32, useMap: true })`
  - `value_content_geolocation_map`

Hence a collision, potentially some bugs there. So I think we should find a different convention.

Perhaps this is enough?

```diff
-_map
+__map
```

Then we know that one underscore symbolizes one object `.` operation, and two underscores (can't do `..` operation in JS) symbolizes something else, kind of metadata.

staltz (changes requested):

index.js
staltz: Oh, I didn't realize that normal prefix indexes had that too! Oops

operators.js
arj03: Good idea

arj03:
Pushed up a new commit. I'm away for the rest of the day, but feel free to merge if you want to try this out in manyverse :) And thanks for the review

staltz (approved):

PR #96, opened 2021-01-23
Only run one index create / update concurrently

arj03:
Converting ssb-about to use jitdb revealed that jitdb doesn't handle concurrent index creates very well. Meaning it will just run as many in parallel as you give it ;) This PR remedies so that only one is running and just calls the waiting after it has finished. Massively faster. There is one caveat this PR handles the simple case of multiple `and(vote())`. But doesn't handle two concurrent requst like `and(author(), isPublic())` and `and(author())`. I don't think this will be a big problem in real life and I value the simplicity of this approach.

Not exactly sure how to write a good test for this. I testing this manually by doing two queries concurrently and with the PR you results in half the time.

### DIFF
--- a/index.js
+++ b/index.js
@@ -371,7 +371,7 @@ module.exports = function (log, indexesPath) {
   }
 
   // concurrent index update
-  const waitingIndexUpdate = {}
+  const waitingIndexUpdate = new Map()
 
   function updateIndex(op, cb) {
     const index = indexes[op.data.indexName]
@@ -464,12 +464,14 @@ module.exports = function (log, indexesPath) {
   }
 
   // concurrent index create
-  const waitingIndexCreate = {}
+  const waitingIndexCreate = new Map()
 
   function createIndexes(opsMissingIndexes, cb) {
     const newIndexes = {}
 
-    const waitingKey = opsMissingIndexes.join()
+    const waitingKey = opsMissingIndexes
+      .map((op) => op.data.indexName)
+      .join('|')
     if (waitingIndexCreate[waitingKey]) {
       waitingIndexCreate[waitingKey].push(cb)
       return // wait for other index create
--- a/index.js
+++ b/index.js
@@ -377,10 +377,10 @@ module.exports = function (log, indexesPath) {
     const index = indexes[op.data.indexName]
 
     const waitingKey = op.data.indexName
-    if (waitingIndexUpdate[waitingKey]) {
-      waitingIndexUpdate[waitingKey].push(cb)
+    if (waitingIndexUpdate.has(waitingKey)) {
+      waitingIndexUpdate.get(waitingKey).push(cb)
       return // wait for other index update
-    } else waitingIndexUpdate[waitingKey] = []
+    } else waitingIndexUpdate.set(waitingKey, [])
 
     // find the next possible seq
     let seq = 0
@@ -455,8 +455,8 @@ module.exports = function (log, indexesPath) {
           else saveIndex(op.data.indexName, index)
         }
 
-        waitingIndexUpdate[waitingKey].forEach((cb) => cb())
-        delete waitingIndexUpdate[waitingKey]
+        waitingIndexUpdate.get(waitingKey).forEach((cb) => cb())
+        waitingIndexUpdate.delete(waitingKey)
 
         cb()
       },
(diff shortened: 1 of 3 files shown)


staltz:
Will review on Monday

staltz:
@arj03 I understood this to be "Only run one log scan at a time, for indexing", right?

I tried adding this benchmark, to see if it would capture the performance boost:

```js
test('query a huge index twice concurrently', (t) => {
  db.onReady(() => {
    const done = multicb({ pluck: 1 })
    const start = Date.now()

    query(
      fromDB(db),
      toCallback(done())
    )

    query(
      fromDB(db),
      toCallback(done())
    )

    done((err) => {
      if (err) t.fail(err)
      const duration = Date.now() - start
      t.pass(`duration: ${duration}ms`)
      fs.appendFileSync(
        reportPath,
        `| Query 1 big index twice concurrently | ${duration}ms |\n`
      )
      t.end()
    })
  })
})
```

Does this accurately reflect what you meant? It doesn't seem to run faster with the PR. A single of these queries is roughly `224ms`, and then running them concurrently is `466ms` with or without this PR.

I have some comments about the code (some possible corner case bugs), but first I'd like to prove on my computer that this gives performance benefits.

arj03:
Strange, are you sure that is creating the indexes? 200ms sounds very low, so it seems more like it is just doing 2 queries on an existing index. But yeah that is a good idea to test it that way. Will give it a try.

staltz:
Aaah, so yeah the indexes are not being *created*, because this is the 3rd run of the query. The 1st run created the index. Will adapt it, just a second.

staltz:
Before PR: `379ms`
After PR: `264ms`

index.js
```diff
@@ -371,9 +371,18 @@ module.exports = function (log, indexesPath) {
     newIndexes[indexName].bitset.add(seq)
   }
 
+  // concurrent index update
+  const waitingIndexUpdate = {}
```
staltz, line 375: `Map` is more suitable (also for performance) if you're adding and `delete`ing frequently

index.js
```diff
@@ -447,13 +456,26 @@ module.exports = function (log, indexesPath) {
...
     })
   }
 
+  // concurrent index create
+  const waitingIndexCreate = {}
+
   function createIndexes(opsMissingIndexes, cb) {
     const newIndexes = {}
+
+    const waitingKey = opsMissingIndexes.join()
```
staltz, line 473: I think this should be `opsMissingIndexes.map(op => op.data.indexName).join()`, right? Otherwise we get `[Object],[Object]` strings

staltz (changes requested):

index.js
staltz: and maybe with a separator, like `join('|')` to make sure we don't do `value_content_x` + `value_content_y` = `value_content_xvalue_content_y`

arj03:
Thanks for the review. Really good points. Will push up a fix in a minute, I found that there were some missing index names in the prefix test ;-)

arj03:
Sorry for the git confusion. Should be there now.

index.js
```diff
@@ -370,9 +370,18 @@ module.exports = function (log, indexesPath) {
...
   }
 
+  // concurrent index update
+  const waitingIndexUpdate = new Map()
+
   function updateIndex(op, cb) {
     const index = indexes[op.data.indexName]
 
+    const waitingKey = op.data.indexName
+    if (waitingIndexUpdate[waitingKey]) {
```
staltz, line 380:
```js
if (waitingIndexUpdate.has(waitingKey)) {
  waitingIndexUpdate.get(waitingKey).push(cb)
  return // wait for other index update
} else waitingIndexUpdate.set(waitingKey, [])
```

index.js
```diff
@@ -446,13 +455,28 @@ module.exports = function (log, indexesPath) {
           else saveIndex(op.data.indexName, index)
         }
 
+        waitingIndexUpdate[waitingKey].forEach((cb) => cb())
+        delete waitingIndexUpdate[waitingKey]
```
staltz, line 459:
```js
waitingIndexUpdate.get(waitingKey).forEach((cb) => cb())
waitingIndexUpdate.delete(waitingKey)
```

More about Maps here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

staltz (changes requested):

index.js
arj03: Right :facepalm:

staltz (approved):

staltz:
Thanks for working out the nitpicks :)

arj03:
Thanks for your patience :)

PR #97, opened 2021-01-25
Add prefixOffset option

arj03:
Makes it possible specifiy where in the target the prefix buffer should be constructed from. 

Some motivation [here](https://github.com/ssb-ngi-pointer/ssb-db2/pull/137), it will make some certain prefix indexes much faster.

### DIFF
--- a/index.js
+++ b/index.js
@@ -302,13 +302,13 @@ module.exports = function (log, indexesPath) {
     }
   }
 
-  function safeReadUint32(buf, offset) {
+  function safeReadUint32(buf, prefixOffset = 0) {
     if (buf.length < 4) {
       const bigger = Buffer.alloc(4)
       buf.copy(bigger)
-      return bigger.readUInt32LE(offset)
+      return bigger.readUInt32LE(prefixOffset)
     } else {
-      return buf.readUInt32LE(offset)
+      return buf.readUInt32LE(prefixOffset)
     }
   }
 
@@ -324,11 +324,10 @@ module.exports = function (log, indexesPath) {
       const fieldStart = opData.seek(buffer)
       if (~fieldStart) {
         const buf = bipf.slice(buffer, fieldStart)
-        const offset = opData.prefixOffset ? opData.prefixOffset : 0
         addToPrefixMap(
           index.map,
           seq,
-          buf.length ? safeReadUint32(buf, offset) : 0
+          buf.length ? safeReadUint32(buf, opData.prefixOffset) : 0
         )
       }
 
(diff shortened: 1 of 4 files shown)


arj03:
Odd, that test failure is unrelated to this change.

arj03:
Seems to be a problem in the test runner...

staltz:
Yeah, seems unrelated. But here's a proposed fix:

```diff
-  state = validate.appendNew(state, null, keys, msg1, Date.now() + 3000)
-  state = validate.appendNew(state, null, keys, msg2, Date.now() + 2000)
-  state = validate.appendNew(state, null, keys, msg3, Date.now() + 1000)
+  const now = Date.now()
+  state = validate.appendNew(state, null, keys, msg1, now + 3000)
+  state = validate.appendNew(state, null, keys, msg2, now + 2000)
+  state = validate.appendNew(state, null, keys, msg3, now + 1000)
```

index.js
```diff
@@ -324,7 +324,12 @@ module.exports = function (log, indexesPath) {
       const fieldStart = opData.seek(buffer)
       if (~fieldStart) {
         const buf = bipf.slice(buffer, fieldStart)
-        addToPrefixMap(index.map, seq, buf.length ? safeReadUint32(buf) : 0)
+        const offset = opData.prefixOffset ? opData.prefixOffset : 0
```
staltz, line 327: :warning: `offset` is a name already taken by the argument to this function. We should use something like `prefixOffset` instead. Preferably also in the arg name of `safeReadUint32` also to make it easier for people to read the code (suppose they are doing a grep for "offset")

arj03:
I think the problem is bigger than that. Working on a fix in a different branch

index.js
```diff
@@ -339,7 +344,8 @@ module.exports = function (log, indexesPath) {
       const fieldStart = opData.seek(buffer)
       if (~fieldStart) {
         const buf = bipf.slice(buffer, fieldStart)
-        index.tarr[seq] = buf.length ? safeReadUint32(buf) : 0
+        const offset = opData.prefixOffset ? opData.prefixOffset : 0
```
staltz, line 347: Same problem with the name `offset`

index.js
```diff
@@ -302,13 +302,13 @@ module.exports = function (log, indexesPath) {
     }
   }
 
-  function safeReadUint32(buf) {
+  function safeReadUint32(buf, offset) {
```
staltz, line 305: I'd rather put the default `0` in the function argument, with

```js
function safeReadUint32(buf, offset = 0) {
```

This is an ES6 feature supported "everywhere modern", even in IE11: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

index.js
staltz: With my suggestion of default function arg above in `safeReadUint32`, we can just directly pass `opData.prefixOffset` as argument to `safeReadUint32`

staltz (changes requested):

index.js
arj03: Nice, I really those default arguments. They were really useful in c++.

arj03:
I'll do a PR against db2 when this is ready for key & others that might have use of this :)

staltz (approved):
Thanks!

PR #99, opened 2021-01-25
refactor updatePrefixMapIndex() a bit

staltz:
Just a tiny refactor. Basically, if `buf.length` is empty, don't even call `addToPrefixMap()`.


arj03:
nice

arj03 (approved):

PR #100, opened 2021-01-26
implement status() Obv API

staltz:
For https://github.com/ssb-ngi-pointer/ssb-db2/issues/139

Here's the before and after of perf benchmarks on my computer. It seems initial index creation is impacted slightly, barely distinguishable from the standard deviation. Other numbers seem fine.

```diff
  core indexes
-   ✔ duration: 9ms
+   ✔ duration: 13ms

  query one huge index (first run)
-   ✔ duration: 671ms
+   ✔ duration: 684ms

  query one huge index (second run)
-   ✔ duration: 238ms
+   ✔ duration: 207ms

  create an index twice concurrently
-   ✔ duration: 256ms
+   ✔ duration: 251ms

  query three indexes (first run)
-   ✔ duration: 388ms
+   ✔ duration: 423ms

  query three indexes (second run)
-   ✔ duration: 204ms
+   ✔ duration: 198ms

  paginate one huge index
-   ✔ duration: 197ms
+   ✔ duration: 179ms

  query a prefix map (first run)
-   ✔ duration: 174ms
+   ✔ duration: 180ms

  query a prefix map (second run)
-   ✔ duration: 5ms
+   ✔ duration: 5ms

  total:     9
  passing:   9
  duration:  2.4s
```


staltz:
The status timer works roughly like this:

- If we called `status.batchUpdate` and the timer isn't running, emit `obv` and start the timer
- If `status.batchUpdate` has been called since last time the timer triggered, emit `obv`
- If `status.batchUpdate` has *NOT* been called since last time the timer triggered, stop the timer

So it's kind of like dynamic polling, it knows when to stop polling.

staltz:
I had both `status.update()` and `status.batchUpdate()` but only the latter was used, so I did this: https://github.com/ssb-ngi-pointer/jitdb/pull/100/commits/c4dea6a2e7fbac4959c96f5a197521de6cf65d6d

Kept the commit in case one day we want to restore `status.update()`.

arj03:
The implementation looks ok. Can we do a seq % 1000 === 0 then status.batchUpdate(indexes, indexNamesForStatus) inside the write loops?

staltz:
I originally had that (as well as another check for `Date.now()`) inside `status.batchUpdate()` but it didn't seem to serve any purpose. I can add it in the push-stream `write` loops, but isn't `1000` too big? Do you have a rough estimate how long it takes to log scan 1000 records? If it's more than 1000ms, we might have problems with the status timer turning off sometimes.

arj03:
With 1000 it should roughly call every 4ms. Would be much more confortable with that in. Otherwise I need to do some testing here to see.

staltz:
I tested it out here, a `log.stream` in db2 (heavier, has decrypt) takes approx 10ms to scan 1000 records. In the worst case it takes 50ms, best case it takes 2ms.

So yes, I'll put that `%` in. :)

arj03 (approved):

staltz:
Lets wait to see the github bot bench results.

Issue #102, opened 2021-01-27
status shows numbers only for numbers that are "moving"


staltz:
I had an idea that can help fix this: https://github.com/ssb-ngi-pointer/ssb-db2/pull/144#issuecomment-768216552

Basically: only report numbers that are "moving", i.e. their indexes are being updated or created. When something else is being updated, delete those numbers from the stats object. Maybe instead of brutally deleting them from the object, we can have two sections `stats.active` and `stats.inactive`.

I'm not entirely sure if this will work but I'll try something.

PR #103, opened 2021-01-27
prune old status numbers

staltz:
When emitting on the `status` obv, check which stats were updated "long ago" and omit them. For issue #102

### DIFF
--- a/status.js
+++ b/status.js
@@ -2,11 +2,23 @@ const Obv = require('obz')
 
 module.exports = function Status() {
   const indexesStatus = {}
+  const indexesLastTime = {}
   const obv = Obv()
   obv.set(indexesStatus)
+  const EMIT_INTERVAL = 1000 // ms
+  const PRUNE_INTERVAL = 2000 // ms
   let i = 0
   let iTimer = 0
   let timer = null
+
+  function pruneStatus() {
+    const now = Date.now()
+    for (const indexName in indexesStatus) {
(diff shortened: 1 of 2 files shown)


status.js
```diff
@@ -2,12 +2,24 @@ const Obv = require('obz')
 
 module.exports = function Status() {
   const indexesStatus = {}
+  const indexesLastTime = {}
```
arj03, line 5: You said in an earlier PR that maps would be better when deleting. Should we do that here as well? Otherwise this looks good.

staltz: Yeah I was thinking about that. If we use a Map, we have to convert it to an object every time we emit the obv. I thought that maybe that's wasteful (object allocation for every obv emit). Gladly the obv emit doesn't happen frequently, so here it was a choice RAM or CPU, and I don't think the CPU will be significantly hurt.

arj03: Alright, lets leave it like this then. Don't think it will be a problem in practise :)

arj03 (approved):

status.js
staltz: Hmm, maybe it's fine if the obv emits Maps, because this is not crossing any muxrpc boundary

staltz: Yeah, bigger fish to catch, but this is up for discussion in the future.

arj03: It's up to you, I'm fine either way.

PR #106, opened 2021-01-28
Better query debug

### LINKED ISSUE #101: Better query debug
So we have these [debugs](https://github.com/ssb-ngi-pointer/jitdb/blob/881eca99abb7e7a60cd9f5a4eb15dd762f1b3cda/index.js#L985) for querys but you sort of have to enable all of jitdb debug to see them. Maybe consider using a jitdb:query namespace for these so you can just enable those. Also make them print the query in some nice to read single line format so it's easy to check. Inspired by debug toolbar in django.

arj03:
This should fix #101. With this you can now do DEBUG="jitdb:query" and get results like:

```
jitdb:query all(AND(AND(type(post),author(@FitmYfxcw)),GT(1))): 0ms, total messages: 1 +261ms
jitdb:query all(type(post)): 0ms, total messages: 1 +260ms
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,6 +8,7 @@ const TypedFastBitSet = require('typedfastbitset')
 const bsb = require('binary-search-bounds')
 const multicb = require('multicb')
 const debug = require('debug')('jitdb')
+const debugQuery = debug.extend('query')
 const Status = require('./status')
 const {
   saveTypedArrayFile,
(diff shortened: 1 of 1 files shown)


staltz (approved):
Boss!

PS: either this PR or #107 need to be updated so that debugQuery works with count()

index.js
```diff
@@ -982,8 +1022,10 @@ module.exports = function (log, indexesPath) {
           (err, answer) => {
             if (err) cb(err)
             else {
-              debug(
-                `paginate(): ${answer.duration}ms, total messages: ${answer.total}`
+              debugQuery(
+                `paginate(${getNameFromOperation(operation)}): ${
+                  answer.duration
+                }ms, total messages: ${answer.total}`
```
staltz, line 1028: Oh, one tiny optimization is to do `debugQuery.enabled && debugQuery(.....)` which means that `getNameFromOperation` would not be called in vain when DEBUG isn't used.

arj03:
Let me just update that and merge, then you can update the other PR

PR #108, opened 2021-01-28
Few fixes

staltz:
Two problems that I'm trying to solve:

(A) the durations in `paginate()` and `all()` seemed to be wrong because they only measured `getMessagesFromBitsetSlice`, not including `executeOperation`. I fixed that

(B) In ssb-db2, we attach `meta.db2`, alongside the existing `meta.db`. I think future developers might run into some WTFs as they try to understand what is "db" when it can refer to ssb-db, jitdb, or ssb-db2 (e.g. `sbot.db.get`). So to make things easier, I decided to not call jitdb "db". So I made a rename. :warning: This is a breaking change, and we should update ssb-db2 ASAP.

### DIFF
--- a/index.js
+++ b/index.js
@@ -983,7 +983,6 @@ module.exports = function (log, indexesPath) {
     cb
   ) {
     seq = seq || 0
-    const start = Date.now()
 
     const sorted = sortedByTimestamp(bitset, descending)
     const sliced =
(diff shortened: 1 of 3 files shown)


arj03:
Looks great. Better to fix now than later.

arj03 (approved):

PR #111, opened 2021-01-29
debugQuery also live queries

staltz:


arj03:
I guess we can do that, but it's a bit different from the other queries in that the query doesn't really end. Anyway if this is useful then lets merge.

staltz:
I'm trying to debug some live-related bugs. For instance:

- Load up the app, see the posts
- Press "like" on one post
- No like appears on the post
- Go to raw log screen
- See all the published msgs, there is no vote in sight

But if I restart the app, then the likes appear. (Note, I'm publishing all msgs via ssb-db1 and then migrate runs)

staltz:
Lets leave this PR open for now, I'll let you know if I'm really confident about it.

staltz:
Yeah I think this is useful, it helped me know when the response to a live query finally came through. It doesn't give duration stats, but that's not the main info I need out of these queries, mostly I need to know the shape of the `op` and the amount of msgs in the result

staltz:
@arj03 Can haz merge?

arj03 (approved):

arj03:
Sure, go ahead :)

PR #112, opened 2021-01-31
Concurrent index load

arj03:
There was another concurrent loading problem. This one in index load. This PR first fixes the problem, then adds a benchmark (goes from 900ms to 650ms) and lastly refactors the waiting code as it was duplicated.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -8,6 +8,7 @@ const rimraf = require('rimraf')
 const mkdirp = require('mkdirp')
 const multicb = require('multicb')
 const ssbKeys = require('ssb-keys')
+const TypedFastBitSet = require('typedfastbitset')
 const JITDB = require('../index')
 const {
   query,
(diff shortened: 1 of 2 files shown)


benchmark/index.js
```diff
@@ -236,6 +238,60 @@ test('query three indexes (second run)', (t) => {
...
+  db.onReady(() => {
+    const done = multicb({ pluck: 1 })
+    const start = Date.now()
+
+    db.indexes['type_about'] = {
+      offset: 0,
+      bitset: new TypedFastBitSet(),
+      lazy: true,
+      filepath: path.join(indexesDir, 'type_about.index'),
+    }
```
staltz, line 254: Is this index going to be loaded by the queries below? It's a lazy index, and we don't seem to be querying `type_about`...

index.js
```diff
@@ -604,15 +610,19 @@ module.exports = function (log, indexesPath) {
...
         cb()
       },
     })
   }
 
+  // concurrent index load
+  const waitingIndexLoad = new Map()
+
   function loadLazyIndex(indexName, cb) {
+    if (onlyOneIndexAtATime(waitingIndexLoad, indexName, cb)) return
```
staltz, line 624: I understand the intent with this code, but I'm confused whether it will actually work. Basically, we're putting this `cb` here in the Map, so it can be called later. And indeed in line 638 it will be called. But the `cb` in this case doesn't actually perform "index load tasks" such as lines 632–636, instead, the `cb` just informs that the loadLazyIndex is "done".

benchmark/index.js
arj03: Whoops fixed.

staltz (approved):

PR #113, opened 2021-02-04
operators should pass prefixOffset onwards

staltz:
I was doing `cat value_author.32prefix | xxd -c 8` on mobile and I kept seeing those `@` sigil in front and thinking "what on Earth, I have the most recent version of db2". Then I looked into the code in JITDB and :sweat_smile:

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -190,8 +190,20 @@ test('query three indexes (first run)', (t) => {
       fromDB(db),
       or(
         and(equal(seekType, 'contact', { indexType: 'type' })),
-        and(equal(seekAuthor, alice.id, { indexType: 'author', prefix: 32 })),
-        and(equal(seekAuthor, bob.id, { indexType: 'author', prefix: 32 }))
+        and(
+          equal(seekAuthor, alice.id, {
+            indexType: 'author',
+            prefix: 32,
+            prefixOffset: 1,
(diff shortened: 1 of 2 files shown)


staltz:
I'm waiting for the benchmarks (before **and** after) to be published here... should see some boost in the "Query 3 indexes" test.

arj03:
Oh my :facepalm:

arj03 (approved):

staltz:
@arj03 Well, there isn't any noticeable boost in performance (in fact the numbers are slightly larger)... We should still merge, right?

arj03:
How are the numbers locally?

staltz:
Actually, locally as well, not a big diff

arj03:
Strange and the indexes have been recreated right?

staltz:
Yeah I ran `npm run benchmark` not `npm run benchmark-no-create`. I also did `cat value_author.32prefix | xxd -c 8` to confirm

arj03:
Let merge this I think. If we do something like this again lets use the version field :)

staltz:
Yeah I was also wondering about the versioning thing. Gladly, with Manyverse, since we I did the screw up with the log decrypt, the whole `db2` folder gets recreated. But this kind of change would need a index version bump.

PR #120, opened 2021-02-10
better debug() of log.stream cases

staltz:
Previously we had a `debug()` when a log.stream ends. This PR adds a `debug()` when the log.stream starts, too, to identify what the purpose of the log.stream is.

### DIFF
--- a/index.js
+++ b/index.js
@@ -427,12 +427,16 @@ module.exports = function (log, indexesPath) {
     let updatedSeqIndex = false
     let updatedTimestampIndex = false
     let updatedSequenceIndex = false
+    const startSeq = seq
     const start = Date.now()
 
     const indexNeedsUpdate =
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -470,7 +474,11 @@ module.exports = function (log, indexesPath) {
       },
       end: () => {
         const count = seq // incremented at end
-        debug(`time: ${Date.now() - start}ms, total items: ${count}`)
+        debug(
+          `log.stream #${logstreamId} done ${seq - startSeq} records in ${
+            Date.now() - start
+          }ms`
+        )
```
staltz, line 481: Note that previously this was reporting `count` which was wrong because that's not the number of records the log.stream went through, it's just the "last seq" when the stream ended. `createIndexes()` doesn't have this issue, only `updateIndex()`

arj03 (approved):

PR #121, opened 2021-02-15
add tests for chainless-and operator inside deferred

staltz:
1st test, 2nd fix


staltz:
This PR is not required for [db2 177](https://github.com/ssb-ngi-pointer/ssb-db2/pull/177) but will make sure that in case we use operators "incorrectly" like it was before [db2 177](https://github.com/ssb-ngi-pointer/ssb-db2/pull/177), then everything will still work.

arj03 (approved):

PR #127, opened 2021-02-22
Async flumelog is now called async-append-only-log

arj03:
Stumpled upon this while writing some documentation for these modules in dev.scuttlebutt.nz


staltz:
Nice! Another one that can be renamed is the `debug()`

staltz (approved):

PR #130, opened 2021-03-01
add test for negative limit given to paginate()

staltz:
1st test, 2nd fix

### DIFF
--- a/index.js
+++ b/index.js
@@ -1062,7 +1062,7 @@ module.exports = function (log, indexesPath) {
     const resultSize = sorted.size
 
     let sliced
-    if (resultSize === 0) {
+    if (resultSize === 0 || limit < 0) {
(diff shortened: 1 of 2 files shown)


staltz:
Hello CI, where are you

staltz:
@arj03 CI is not running today, not sure why, but if you want you can run tests locally.

arj03 (approved):
This looks good. Thanks, simple fix.

arj03:
While we are at it (in another PR). Should paginate(0) return 0 results instead of all as it currently does?

staltz:
> While we are at it (in another PR). Should paginate(0) return 0 results instead of all as it currently does?

Yes, makes lot of sense. I'm not sure what part of the tests attempt limit=0, but I'll take a look.

PR #131, opened 2021-03-01
paginate() with limit=0 should return zero results

staltz:
Before, we were using falsy values of `limit` to mean "everything". Now, falsy values mean "Infinity", and `db.all` will use `Infinity`. So `limit=0` is not anymore interpreted as falsy, it's interpreted as zero, and the results are thus empty.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1062,7 +1062,7 @@ module.exports = function (log, indexesPath) {
     const resultSize = sorted.size
 
     let sliced
-    if (resultSize === 0 || limit < 0) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1132,7 +1132,7 @@ module.exports = function (log, indexesPath) {
         getMessagesFromBitsetSlice(
           bitset,
           seq,
-          null,
+          Infinity,
```
arj03, line 1135: nice, this makes things more readable as well.

arj03 (approved):

PR #134, opened 2021-03-19
Optimize edge cases for getMessagesFromBitsetSlice

Barbarrosa:
I attempted some optimizations to `getMessagesFromBitsetSlice`.

- Use simpler comparisons where possible
- Prefer `peek` over `kSmallest` when returning a single element, not only for zero offset
  - [The queue lib](https://github.com/lemire/FastPriorityQueue.js/blob/master/FastPriorityQueue.js#L274) seems way less efficient for kSmallest at size one

I'm not terribly familiar w/ this codebase, so let me know if I'm missing something here.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1064,15 +1064,16 @@ module.exports = function (log, indexesPath) {
   ) {
     seq = seq || 0
 
+    const returnTrue = () => true
     let sorted = sortedByTimestamp(bitset, descending)
-    const resultSize = sorted.size
+    const total = sorted.size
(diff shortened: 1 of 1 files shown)


Barbarrosa:
The performance benchmarks aren't showing the improvement, so I'm going to close this PR. If the benchmarks show a concrete improvement after changes, then I may reopen it or open another PR.

FYI the benchmark job is failing b/c it requires repo permissions in order to add the PR comment.

PR #137, opened 2021-03-21
Improve Pagination Performance

Barbarrosa:
This MR uses alternative `kSmallest` and `clone` methods for the FastPriorityQueue to speed up `getMessagesFromBitsetSlice`. I also submitted these changes to [lemire/FastPriorityQueue.js](https://github.com/lemire/FastPriorityQueue.js/pull/29).

I don't think I have the right permissions, so the benchmark job will probably fail to add a comment (so you'll need to examine the job to see results). Based on my own testing, this seems to double performance for pagination over small page sizes and somewhat improves performance for large page sizes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "binary-search-bounds": "^2.0.4",
     "bipf": "^1.4.0",
     "debug": "^4.2.0",
-    "fastpriorityqueue": "^0.6.3",
+    "fastpriorityqueue": "^0.7.0",
     "idb-kv-store": "^4.5.0",
     "jsesc": "^3.0.2",
     "mkdirp": "^1.0.4",


Barbarrosa:
The upstream PR was merged, so I updated this PR to use the updated lib instead of reimplementing `kSmallest` and `clone`.

Barbarrosa:
@arj03 Would you please take a look? I don't have permissions to formally request review.

arj03:
This is great, was waiting for upstream to accept changes. Testing this now.

arj03:
Before

```
 paginate big index with small pageSize

    ✔ duration: 19401ms

  paginate big index with big pageSize

    ✔ duration: 339ms
```

After:

```
  paginate big index with small pageSize

    ✔ duration: 4721ms

  paginate big index with big pageSize

    ✔ duration: 257ms
```

staltz:
Thanks @Barbarrosa for this! I love it that there was some back and forth with the solution, and eventually it turned out to be a PR to the dependency, and just a dependency update in our repo. :100:

Barbarrosa:
@staltz You're welcome! I hope this helps with performance in the next Manyverse update.

Issue #138, opened 2021-03-23
"or" operator behavior changes at top level


Barbarrosa:
I was looking over the different operators when I noticed a repeated pattern that seems confusing. Take the example from the readme:
```typescript
query(
  fromDB(db),
  and(slowEqual('value.content.type', 'contact')),
  and(or(slowEqual('value.author', aliceId), slowEqual('value.author', bobId))), // this line
  toCallback((err, msgs) => {
    console.log('There are ' + msgs.length + ' messages')
  })
)
```
I see the `or` operator almost universally wrapped in an `and` operator. When I dug in, I found that the `or` part applied not only to the `or` method's parameters but also to prior operations *when at the top level*. The code below functionally differs from the above, i.e. including non-`contact` objects with matching `author` values:
```typescript
query(
  fromDB(db),
  and(slowEqual('value.content.type', 'contact')),
  or(slowEqual('value.author', aliceId), slowEqual('value.author', bobId)),
  toCallback((err, msgs) => {
    console.log('There are ' + msgs.length + ' messages')
  })
)
```
The below example is functionally identical to the first example, and it shows the behavior does not persist below the top level of the query.
```typescript
query(
  fromDB(db),
  and(
    slowEqual('value.content.type', 'contact'),
    or(
      slowEqual('value.author', aliceId),
      slowEqual('value.author', bobId)
    )
  ),
  toCallback((err, msgs) => {
    console.log('There are ' + msgs.length + ' messages')
  })
)
```

Technically it's a BC break, but it seems like the `or` operator shouldn't apply itself to prior operations at the top level.

staltz:
Yeah, you're right. I think it would be best to recommend using only one "and-or" tree as an argument, i.e. recommend the 3rd example, not the 1st nor the 2nd. It's easier to understand and predict how it'll behave.

Barbarrosa:
I was thinking about doing it automatically - i.e. something like #139. That kind of change may be an issue depending on adoption since it's technically a BC break, but such usages are likely incorrect anyways.

staltz:
@Barbarrosa I'm thinking about this one still, I wouldn't want to merge something that changes semantics, then revert it, so I'd like to be very sure about this one before merging.

Now that I take a second look, I think the 2nd code example is actually functionally correct and we'd like to support it. Technically, we should always recommend the 3rd example (I think the readme examples need to be updated), but there's yet another use case to consider: some queries can be split into reusable parts, e.g.:

```js
[...]

PR #139, opened 2021-03-24
Make or operator consistent at top level vs nested - see #138

Barbarrosa:
This wraps the "or" operator an an "and" operator whenever it's included at the top level of a query.


Barbarrosa:
Given that #138 may lead to a more complete solution to this & similar API usability issues, I'm closing this PR.

PR #140, opened 2021-03-27
Update FastPriorityQueue with removeMany bugfix

Barbarrosa:
FYI some of the npm packages for `fastpriorityqueue` appear to be versioned incorrectly. This actually includes the enhancement from version `0.7.0` even though it appears to have a lower version number.

You can reference lemire/FastPriorityQueue.js#30 and lemire/FastPriorityQueue.js#31 for exactly what's fixed here. I believe your current usage of `removeMany` will be unaffected by the bug, but the fixed version may perform worse in your benchmarks. It may prevent bugs if someone updates that area of the code.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1075,7 +1075,9 @@ module.exports = function (log, indexesPath) {
     } else {
       if (seq > 0) {
         sorted = sorted.clone()
-        sorted.removeMany(() => true, seq)
+        for (let j = 0; j < seq && !sorted.isEmpty(); j++) {
+          sorted.poll()
+        }
       }
       sliced = sorted.kSmallest(limit || Infinity)
     }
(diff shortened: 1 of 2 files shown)


Barbarrosa:
@arj03 I switched from `removeMany` to a simple poll loop because the fixed version performs worse in your benchmarks.

arj03:
Thanks @Barbarrosa 

This does seem faster, mostly in the `paginate big index with small pageSize` benchmark. I think I'll wait and see if he doesn't release a 0.7.1.

staltz:
> I think I'll wait and see if he doesn't release a 0.7.1 before merging.

:+1: 


Another note: even if benchmark results don't appear as a comment here, you can see them over there:

Screenshot from 2021-03-27 10-46-08

Barbarrosa:
@arj03 The upstream maintainer published version 0.7.1 and I updated this PR with the new version.

arj03:
Thanks :)

PR #141, opened 2021-04-03
add where() operator and change and() and or()

staltz:
**BREAKING CHANGE**

Related to: #138 and #139

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ query the database. You can load these operators from
 ```js
 const Log = require('async-append-only-log')
 const JITDB = require('jitdb')
-const { query, fromDB, and, slowEqual, toCallback } = require('jitdb/operators')
+const {query, fromDB, where, slowEqual, toCallback} = require('jitdb/operators')
 
 const raf = Log('/home/me/path/to/async-log', {
   blockSize: 64 * 1024,
@@ -70,7 +70,7 @@ const db = JITDB(raf, '/home/me/path/to/indexes')
 db.onReady(() => {
   query(
     fromDB(db),
-    and(slowEqual('value.content.type', 'post')),
+    where(slowEqual('value.content.type', 'post')),
     toCallback((err, msgs) => {
       console.log(msgs)
     })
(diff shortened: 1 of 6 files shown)


staltz:
cc @Barbarrosa @mixmix

README.md
```diff
@@ -159,11 +165,13 @@ query(
...
-    or(
-      equal(seekAuthor, aliceId, { indexType: 'author' }),
-      equal(seekAuthor, bobId, { indexType: 'author' })
+  where(
+    and(
+      equal(seekType, 'contact', 'type')
+      or(
+        equal(seekAuthor, aliceId, { indexType: 'author' }),
+        equal(seekAuthor, bobId, { indexType: 'author' })
+      )
```
arj03, line 174: nice

operators.js
```diff
@@ -263,45 +263,18 @@ function not(ops) {
...
-          }
-        : rhs[0]
-    if (ops) copyMeta(ops, res)
+  return { type: 'OR', data: args }
+}
+
+function where(nextOp) {
+  return (prevOp) => {
+    if (!nextOp) return prevOp
+    const res = prevOp.type ? { type: 'AND', data: [prevOp, nextOp] } : nextOp
```
arj03, line 276: This is actually a bit easier to read than the others. If I read this correctly it is: support null, support normal case, support where outside chain.

test/operators.js
```diff
@@ -167,11 +159,11 @@ prepareAndRunTest('and() ignores non-function arguments', dir, (t, db, raf) => {
...
 
-prepareAndRunTest('or() ignores non-function arguments', dir, (t, db, raf) => {
+prepareAndRunTest('where() takes only one arguments', dir, (t, db, raf) => {
   const queryTree = query(
     fromDB(db),
-    or(
-      null,
+    where(
+      equal(helpers.seekType, 'contact', { indexType: 'type', indexAll: true }),
       equal(helpers.seekType, 'post', { indexType: 'type', indexAll: true })
```
arj03, line 167: in this case the second argument is ignored, right?

operators.js
staltz: Yes, overall this file had a net negative git diff. You guessed correctly, but the 3rd case is "support `where` when it's the first thing right after `fromDB(db)`".

test/operators.js
arj03: I wonder if it would be better to throw an error

operators.js
arj03: Yes the negative diff in the amount of code is really nice and a good sign that this is a better abstraction.

test/operators.js
staltz: Hmm, yeah we can make `where` throw.

arj03 (approved):

staltz:
Before merging, I'll try this out in ssb-db2 and see if it all works.

mixmix:
I've just read the readme. Really sweet seeing these changes in response to that thread.
I appreciate how much work changing things like this around can be

# 🌱🍄🐉

staltz:
Thanks @mixmix 

Ok, @arj03 I tested ssb-db2 and it works. I'll merge and make a new version later. It  will have to be a major version.

Barbarrosa:
@staltz @arj03 Thanks, this makes JITDB queries far more predictable.

PR #142, opened 2021-04-06
Delay checking prefix index until we need the result

arj03:
I was looking at the [performance of getting messages for a specific author](https://github.com/arj03/ssb-sync-test) and remembered there was one thing that I wanted to fix about prefix indexes, namely that it should be a bit more lazy about checking the messages it needs. Basically right now if you do query of: *give me the latest 10 messages from a specific author* it has to check all messages for that author to make sure that they match the prefix before it can sorted and results returned.

:warning: this changes the core structure of queries so should be reviewed carefully :warning: 

I made several rounds of this and the tests + benchmarks we have were quite useful in making sure that the code still works as before. I added an extra test to make sure that it works like I expected. Also this PR fixes a bug in the benchmarking code where we did not wait for an index to be written to disc before continuing and the bug was masked by the old code having to get a ton of messages from disc.

Benchmark results:

Before

```
db query warmup (10319): 2.985s
db query author all (8032): 959.953ms
db query author post msgs (2095): 617.328ms
db query author 10 msgs: 198.095ms
```

After:

```
db query warmup (10319): 1.013s
db query author all (8032): 760.335ms
db query author post msgs (2095): 236.625ms
db query author 10 msgs: 26.373ms
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -877,6 +877,7 @@ module.exports = function (log, indexesPath) {
   }
 
   function mergeFilters(filters1, filters2) {
+    if (!filters1 && !filters2) return null
     const filters = filters1 || new Map()
     if (!filters2) return filters
     for (let seq of filters2.keys()) {
--- a/index.js
+++ b/index.js
@@ -881,7 +881,7 @@ module.exports = function (log, indexesPath) {
     else if (filters1 && !filters2) return filters1
     else if (!filters1 && filters2) return filters2
     else {
-      const filters = filters1
+      const filters = new Map(filters1)
       for (let seq of filters2.keys()) {
         const f1 = filters1.get(seq) || []
         const f2 = filters2.get(seq)
--- a/index.js
+++ b/index.js
@@ -878,14 +878,17 @@ module.exports = function (log, indexesPath) {
 
   function mergeFilters(filters1, filters2) {
     if (!filters1 && !filters2) return null
-    const filters = filters1 || new Map()
-    if (!filters2) return filters
-    for (let seq of filters2.keys()) {
-      const f1 = filters.get(seq) || []
-      const f2 = filters2.get(seq)
-      filters.set(seq, [...f1, ...f2])
+    else if (filters1 && !filters2) return filters1
+    else if (!filters1 && filters2) return filters2
+    else {
+      const filters = filters1
+      for (let seq of filters2.keys()) {
+        const f1 = filters1.get(seq) || []
+        const f2 = filters2.get(seq)
+        filters.set(seq, [...f1, ...f2])
+      }
+      return filters
     }
-    return filters
   }
 
   function getBitsetForOperation(op, cb) {
--- a/index.js
+++ b/index.js
@@ -1071,12 +1071,14 @@ module.exports = function (log, indexesPath) {
   }
 
   function filterRecord(seq, filters, recBufferCache, cb) {
+    const seqFilters = filters.get(seq)
+    if (!seqFilters) return cb(null, seq)
+
     getRecord(seq, (err, record) => {
       if (err) return cb(err)
 
       const recBuffer = record.value
       let ok = true
-      const seqFilters = filters.get(seq)
       if (seqFilters) ok = seqFilters.every((filter) => filter(recBuffer))
 
       if (ok) {
--- a/index.js
+++ b/index.js
@@ -876,6 +876,17 @@ module.exports = function (log, indexesPath) {
     }
   }
 
+  function mergeFilters(filters1, filters2) {
+    const filters = filters1 || new Map()
+    if (!filters2) return filters
+    for (let seq of filters2.keys()) {
+      const f1 = filters.get(seq) || []
+      const f2 = filters2.get(seq)
+      filters.set(seq, [...f1, ...f2])
+    }
+    return filters
+  }
+
   function getBitsetForOperation(op, cb) {
     if (op.type === 'EQUAL' || op.type === 'INCLUDES') {
       if (op.data.prefix) {
@@ -910,12 +921,7 @@ module.exports = function (log, indexesPath) {
 
       getBitsetForOperation(op.data[0], (op1, filters1) => {
         getBitsetForOperation(op.data[1], (op2, filters2) => {
-          const filters = filters1 || new Map()
-          if (filters2)
-            for (let k of filters2.keys())
-              filters.set(k, [...(filters.get(k) || []), ...filters2.get(k)])
-
-          cb(op1.new_intersection(op2), filters)
+          cb(op1.new_intersection(op2), mergeFilters(filters1, filters2))
         })
       })
     } else if (op.type === 'OR') {
@@ -923,12 +929,7 @@ module.exports = function (log, indexesPath) {
 
       getBitsetForOperation(op.data[0], (op1, filters1) => {
         getBitsetForOperation(op.data[1], (op2, filters2) => {
-          const filters = filters1 || new Map()
-          if (filters2)
-            for (let k of filters2.keys())
-              filters.set(k, [...(filters.get(k) || []), ...filters2.get(k)])
-
-          cb(op1.new_union(op2), filters)
+          cb(op1.new_union(op2), mergeFilters(filters1, filters2))
         })
       })
     } else if (op.type === 'NOT') {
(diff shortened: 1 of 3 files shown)


staltz:
I'm still taking a look at this and wrapping my head around it, but one feedback is that we should probably use `Map` for the filters dict, it might improve performance a lot. To iterate over the keys, we should probably do `for (let key of map.keys())`

See these benchmarks I made: https://jsbench.me/2pkn5xz28v/1 (Left is Firefox, right is Chrome V8 and also what runs in Node.js)

Screenshot from 2021-04-06 14-32-57

index.js
```diff
@@ -1012,7 +1015,8 @@ module.exports = function (log, indexesPath) {
     else return true
   }
 
-  function getMessage(seq, cb) {
+  function getMessage(seq, valueCache, cb) {
+    if (valueCache[seq]) return cb(null, bipf.decode(valueCache[seq], 0))
```
staltz, line 1019: What's the performance improvements this cache adds? I thought we could rely on the cache already inside log.get: https://github.com/ssb-ngi-pointer/async-append-only-log/blob/a75f71485afbbda627fb20a6ef8aafd133c3f85f/index.js#L119

arj03: Without the cache:

```
db query warmup: 1.802s
db query author all: 693.186ms
db query author post msgs: 164.51ms
db query author 10 msgs: 19.311ms
```

With:

```
db query warmup: 611.244ms
db query author all: 495.455ms
db query author post msgs: 153.325ms
db query author 10 msgs: 18.665ms
```

I was actually puzzled why this PR was so much better at startup performance, this seems to somewhat answer that.

This is running with power connected, so it is a bit faster than the numbers I posted earlier. That is why I included both.

arj03:
Ohh, the map was a good one:

MAP:

```
db query warmup: 576.999ms
db query author all: 491.68ms
db query author post msgs: 139.234ms
db query author 10 msgs: 13.678ms
```

PR:

```
db query warmup: 611.244ms
db query author all: 495.455ms
db query author post msgs: 153.325ms
db query author 10 msgs: 18.665ms
```

index.js
staltz: Ok, I just wanted to be sure we don't have extra caches (they have a memory hit after all). We could also try `Map` cache instead of object cache.

index.js
```diff
@@ -940,7 +945,7 @@ module.exports = function (log, indexesPath) {
 
   function executeOperation(operation, cb) {
     updateCacheWithLog()
-    if (bitsetCache.has(operation)) return cb(bitsetCache.get(operation))
+    if (bitsetCache.has(operation)) return cb(...bitsetCache.get(operation))
```
staltz, line 948: One thing I'm concerned is whether the `filters` Map will be reused, because when there are AND or OR operations, the 1st `filters` Map gets *mutated*, and in this `bitsetCache` we are storing that mutated Map.

It could be we're totally safe, but I'm just trying to think if we have that guarantee or not.

arj03: I had to double check but I think we should be safe. Basically as long as the log does not change (which the cache already has a check for), then for a given operation the bitset + filters should be exactly the same. The only thing that could change is the seq or limit and for those we can reuse the bitset and filters as they are not mutated in `getMessagesFromBitsetSlice`.

arj03: I take that back. If you do have a query like `or(and(type('post'), author('a')), and(type('vote'), author('a')))` then that could potentially have a problem as you describe. I tested just creating a new map when merging and could not measure any perf diff, so better safe than sorry.

Barbarrosa:
I was curious, so I ran the benchmark from #143 both before and after this change. Here are the results I got.  I read this as showing much leaner memory usage, but with some cost to speed.

I ran this after incorporating something similar to the benchmark bugfix in #142.

## Before

| Part | Speed | Heap Change | Samples |
|---|---|---|---|
| Load core indexes | 0.81ms ± 0.01ms | 435 B ± 33290 B | 418 |
| Query 1 big index (1st run) | 628.55ms ± 2.75ms | -241.92 kB ± 17.81 kB | 5 |
| Query 1 big index (2nd run) | 220.56ms ± 10.07ms | -284.24 kB ± 0 kB | 5 |
| Count 1 big index (3rd run) | 0.13ms ± 0ms | -467.72 kB ± 7.16 kB | 5 |
| Create an index twice concurrently | 535.85ms ± 36.27ms | 1.28 MB ± 0.04 MB | 5 |
| Query 3 indexes (1st run) | 648.78ms ± 3.79ms | 2.38 MB ± 0 MB | 5 |
| Query 3 indexes (2nd run) | 181.91ms ± 5.68ms | -65.67 kB ± 0 kB | 5 |
| Load two indexes concurrently | 336.02ms ± 15.4ms | 13.41 MB ± 0 MB | 5 |
| Paginate 20000 msgs with pageSize=5 | 8894.7ms ± 3624.68ms | -101.32 kB ± 0 kB | 5 |
| Paginate 20000 msgs with pageSize=500 | 405.45ms ± 32.54ms | 2.16 kB ± 0 kB | 5 |
| Query a prefix map (1st run) | 177.74ms ± 5.54ms | 5.03 MB ± 0.04 MB | 5 |
| Query a prefix map (2nd run) | 7.55ms ± 0.53ms | 2.49 MB ± 0.03 MB | 5 |

## After

| Part | Speed | Heap Change | Samples |
|---|---|---|---|
| Load core indexes | 0.81ms ± 0.01ms | 452 B ± 9951 B | 410 |
| Query 1 big index (1st run) | 623.05ms ± 5.11ms | -750.39 kB ± 45 kB | 5 |
| Query 1 big index (2nd run) | 228.82ms ± 10.64ms | -6.49 kB ± 42.09 kB | 5 |
| Count 1 big index (3rd run) | 0.13ms ± 0ms | -1.58 MB ± 0 MB | 5 |
| Create an index twice concurrently | 507.61ms ± 5.69ms | -15.51 kB ± 42.33 kB | 5 |
| Query 3 indexes (1st run) | 757.49ms ± 8.14ms | 397.25 kB ± 0 kB | 5 |
| Query 3 indexes (2nd run) | 276.89ms ± 15.39ms | 295.7 kB ± 0 kB | 5 |
| Load two indexes concurrently | 530.35ms ± 10.39ms | -558.9 kB ± 44.37 kB | 5 |
| Paginate 20000 msgs with pageSize=5 | 9009.28ms ± 3519.35ms | -4.29 MB ± 0 MB | 5 |
| Paginate 20000 msgs with pageSize=500 | 409.66ms ± 33.34ms | -488.18 kB ± 35.92 kB | 5 |
| Query a prefix map (1st run) | 177.39ms ± 1.85ms | 44.16 kB ± 0 kB | 5 |
| Query a prefix map (2nd run) | 9.66ms ± 0.37ms | 7.17 MB ± 0.03 MB | 5 |

staltz:
@Barbarrosa Thanks for the useful info. It's a bit puzzling why do the benchmarks not look better with this PR.

> db query warmup (10319): 1.013s
> db query author all (8032): 760.335ms
> db query author post msgs (2095): 236.625ms
> db query author 10 msgs: 26.373ms

@arj03 Could you share the code snippets to produce these author benchmarks?

arj03:
I used server.js from [here](https://github.com/arj03/ssb-sync-test). I can maybe try adding that as a benchmark here.

arj03:
The reason for running them outside was that I wanted to know how it performs on real live data.

arj03:
Added a benchmark where this new code is better. Before it was 58ms, now it is 15ms.

I'll try to see if I can figure out why `Query 3 indexes (2nd run)` is slower. That is the one I'm most worried about.

arj03:
So the problem was that `query 3` basically end up with query that looks like 2000 results that needs to be filter checked and then 20k that don't (the contact type) and before we did the filter check on all of them.

arj03:
Thanks @Barbarrosa, your input was very helpful. Can't wait to have your other PR merged.

staltz (approved):
The callback part is difficult to understand (maintain too), but I'm fine if this is merged now. I could try refactoring it later but I don't want to block this PR. It's great we found a way to optimize even further, I thought we had exhausted all the ideas.

arj03:
Thanks for the help with this @staltz

PR #144, opened 2021-04-07
refactor bitset filters logic

staltz:
@arj03 This is some of my code review for #142 in the form of a PR `faster-prefix2` –> `faster-prefix`.

I think I still am confused with the callback hell around `processResults`. I am tempted to simplify it with a push-stream pipeline, it seems possible. I might do that still this week.

### DIFF
--- a/benchmark/index.js
+++ b/benchmark/index.js
@@ -263,15 +263,11 @@ test('load two indexes concurrently', (t) => {
   const bob = ssbKeys.loadOrCreateSync(path.join(dir, 'secret-b'))
 
   function waitForFile(filename, cb) {
-    fs.exists(filename, function(exists) {
-      if (exists) cb()
-      else setTimeout(() => waitForFile(filename, cb), 250)
-    })
+    if (fs.existsSync(filename)) cb()
+    else setTimeout(waitForFile, 250, filename, cb)
   }
 
   db.onReady(() => {
-    const done = multicb({ pluck: 1 })
-    const start = Date.now()
     const contact_index_filename = path.join(indexesDir, 'type_contact.index')
 
     // we need to wait for the other tests to write the index file
@@ -282,6 +278,9 @@ test('load two indexes concurrently', (t) => {
         lazy: true,
         filepath: contact_index_filename,
       }
+
+      const done = multicb({ pluck: 1 })
+      const start = Date.now()
 
       query(
         fromDB(db),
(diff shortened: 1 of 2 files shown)


benchmark/index.js
```diff
@@ -263,15 +263,11 @@ test('load two indexes concurrently', (t) => {
...
-      if (exists) cb()
-      else setTimeout(() => waitForFile(filename, cb), 250)
-    })
+    if (fs.existsSync(filename)) cb()
+    else setTimeout(waitForFile, 250, filename, cb)
   }
 
   db.onReady(() => {
-    const done = multicb({ pluck: 1 })
-    const start = Date.now()
```
staltz, line 274: We shouldn't trigger the `start` before the `waitForFile` is done, because it uses some 250ms timers.

index.js
```diff
@@ -1060,16 +1067,17 @@ module.exports = function (log, indexesPath) {
     return fpq
   }
 
-  function filterRecords(seq, filters, valueCache, cb) {
+  function filterRecord(seq, filters, recBufferCache, cb) {
```
staltz, line 1070: I thought that `valueCache` is too vague, so used `recBuffer` terminology

index.js
```diff
@@ -1102,18 +1110,18 @@ module.exports = function (log, indexesPath) {
...
+    const recBufferCache = {}
 
-    function processResults(err, seqs, resultSize) {
+    function processResults(seqs, resultSize) {
       push(
         push.values(seqs),
-        push.asyncMap((seq, cb) => {
-          if (onlyOffset) cb(null, indexes['seq'].tarr[seq])
-          else getMessage(seq, valueCache, cb)
+        push.asyncMap((seq, continueCB) => {
```
staltz, line 1122: Renamed this callback to avoid confusion with the outer `cb`

index.js
```diff
@@ -1137,34 +1146,30 @@ module.exports = function (log, indexesPath) {
             startSeq + rawLength,
             limit - seqs.length,
             seqs,
-            (err, seqs) => ensureEnoughResults(err, startSeq + rawLength, seqs)
+            (e, seqs) => ensureEnoughResults(e, startSeq + rawLength, seqs)
           )
-        else processResults(err, seqs, resultSize, cb)
+        else processResults(seqs, resultSize)
```
staltz, line 1151: `processResults` wasn't using the last argument :|

arj03: Ah, that was from when I had those functions a bit further out but realized that I was passing a ton of arguments to each function. Good find.

arj03 (approved):

arj03:
Looks great, thanks. Was thinking of refactoring the mergefilters function so glad to see you did that here.

PR #147, opened 2021-04-10
Skip null arguments in AND, OR operator

arj03:
I was upgrading (surprisingly easy and much better code) ssb-browser-demo to the new where() syntax and noticed that null skipping was not working for AND / OR like it was before.

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -263,11 +263,11 @@ function not(ops) {
 }
 
 function and(...args) {
-  return { type: 'AND', data: args }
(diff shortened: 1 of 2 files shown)


staltz (approved):
I had left that out because I didn't think it would be used anymore (because I thought the use case would be `opts.live ? live({old:false}) : null` which would be outside the where()), but this shows that you had other use cases, so it's good to merge. :)

arj03:
Thanks for the review :)

PR #152, opened 2021-04-15
support and(op, null) and and(null, null)

staltz:
We were supporting `and(op1, op2, null)` but we had failures for `and(op, null)` and `and(null, null)`.

1st test, 2nd fix


arj03 (approved):
nice

Issue #149, opened 2021-04-11
Corrupted indexes


staltz:
This is a placeholder for more details that I'll post later, because I'm on mobile.

We found a bug in Manyverse where opening a thread for msg root X would actually open msg Y, completely unrelated msg. Also we had other spooky issues, like some threads not appearing on the public feed despite them being in the local log, or even when I publish a msg myself, sometimes it didn't show on the public feed.

I confirmed that this was caused by jitdb or ssb-db2 by deleting the indexes folder and letting it reindex. After that, the bugs disappeared.

Gladly, I kept a copy of the corrupted indexes folder. I also kept a copy of the newly built good indexes folder, and both of these should be matching the same log, because I put airplane mode on and had no new replication. This way, we can do a raw diff on the corrupted-vs-good indexes folders and discover what's going on.

If I'd have to guess, I would say there was a bug in atomically and the core indexes like offset and timestamp began accumulating errors as the file got corrupted.

arj03:
Good you saved the indexes. Would be very interested in taking a look.

arj03:
Vote & author indexes are the same. Seq & timestamp are also ok after loading and checking. Post does have 2 differences at index 421 and 422. The rest is the same. I'll do some more debugging and post my tool to look at the files.

arj03:
Some more debugging info mostly for myself:

# value_content_root_.index (isRoot)

```
bitset length diff
missing in corrupt [
  1055962, 1055963, 1055964, 1055965, 1055966,
  1055968, 1055969, 1055972, 1055974, 1055975,
  1055976, 1055977, 1055978, 1055979, 1055980,
  1055981, 1055983, 1055985, 1055987, 1055988,
  1055991, 1055992, 1055993, 1055994, 1055995,
  1055997, 1055998, 1055999, 1056000, 1056001,
  1056002, 1056003, 1056004, 1056005, 1056006,
  1056007, 1056008, 1056010, 1056012, 1056013,
  1056015, 1056016, 1056017, 1056018, 1056020,
  1056022, 1056023, 1056024, 1056026, 1056027,
  1056028, 1056029, 1056030, 1056031, 1056032,
  1056033, 1056034, 1056036, 1056041, 1056043
]
missing in ok [ 13505, 13506, 13507, 13508 ]
```

# value_content_root__map.32prefixmap (hasRoot)

```
missing keys in corrupt []
missing keys in ok []
[...]

PR #153, opened 2021-04-18
Crc

arj03:
This adds a crc32 checks to indexes. This PR depends on [fix concurrency PR in atomic-file-rw](https://github.com/ssb-ngi-pointer/atomic-file-rw/pull/2). The idea is to calculate a checksum of the data before writing and checking on load again. The overhead is not super [bad](https://github.com/arj03/hash-test/) and we'd rather be safe than sorry and run into something [corruption issues](https://github.com/ssb-ngi-pointer/jitdb/issues/149). This PR assumes that previous indexes are ok (crc=0). If one could rather have them recalculated they should be manually deleted, seeing as atomically created a bunch of tmp files a manual process is probably needed anyway.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1007,7 +1007,7 @@ module.exports = function (log, indexesPath) {
         )
       }),
 
-      // some indexes might have failed to load
+      // some lazy indexes may have failed to load
       push.asyncMap((_, next) => {
         opsMissingIdx = detectMissingAndLazyIndexes(operation)[0]
         if (opsMissingIdx.length > 0)
--- a/index.js
+++ b/index.js
@@ -954,16 +954,14 @@ module.exports = function (log, indexesPath) {
     } else console.error('Unknown type', op)
   }
 
-  function detectMissingAndLazyIndexes(operation) {
-    const opsMissingIdx = []
-    const lazyIdxNames = []
+  function detectLazyIndexesUsed(operation) {
+    const results = []
 
     function detectMore(ops) {
       ops.forEach((op) => {
         if (op.type === 'EQUAL' || op.type === 'INCLUDES') {
-          const indexName = op.data.indexName
-          if (!indexes[indexName]) opsMissingIdx.push(op)
-          else if (indexes[indexName].lazy) lazyIdxNames.push(indexName)
+          const name = op.data.indexName
+          if (indexes[name] && indexes[name].lazy) results.push(name)
         } else if (op.type === 'AND' || op.type === 'OR' || op.type === 'NOT')
           detectMore(op.data)
         else if (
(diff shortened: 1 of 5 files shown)


Barbarrosa:
@arj03 I added a PR here for error handling: https://github.com/ssb-ngi-pointer/atomic-file-rw/pull/3

arj03:
Thanks @Barbarrosa. I merged that and released as 0.2.1. Really good with some extra eyes to spot those corner cases.

index.js
```diff
@@ -995,21 +1007,29 @@ module.exports = function (log, indexesPath) {
         )
       }),
 
-      // Create missing indexes, if any
+      // some indexes might have failed to load
```
staltz, line 1010:
```suggestion
      // some lazy indexes may have failed to load
```

index.js
```diff
@@ -995,21 +1007,29 @@ module.exports = function (log, indexesPath) {
...
+      push.asyncMap((_, next) => {
+        opsMissingIdx = detectMissingAndLazyIndexes(operation)[0]
```
staltz, line 1012:
```suggestion
        [opsMissingIdx] = detectMissingAndLazyIndexes(operation)
```

staltz: It's a bit unfortunate that the first array `opsMissingIdx` is calculated but never used, because it will always be re-calculated here. Maybe we could split up `detectMissingAndLazyIndexes` to `detectMissingIndexes` and `detectLazyIndexes`? (I can do that if you want)

arj03: Yeah I was thinking of splitting that function up because we only use missing on second go. Feel free to fix that :)

staltz:
files.js looks good and the tests are very welcome, thanks. I'll make a tiny PR to update `detectMissingAndLazyIndexes`

staltz:
In my tests locally I also get these logs quite often, and they appear also in CI:

```
[Error: ENOENT: no such file or directory, rename '/tmp/jitdb-add/indexesUpdate index/timestamp.index~' -> '/tmp/jitdb-add/indexesUpdate index/timestamp.index'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'rename',
  path: '/tmp/jitdb-add/indexesUpdate index/timestamp.index~',
  dest: '/tmp/jitdb-add/indexesUpdate index/timestamp.index'
}
```

They are not "failing errors" but they are polluting the log a bit.

arj03:
That is because you are running an old version of atomic-file-rw?

staltz:
Ah right, this PR is pointing to 0.1.0 and it kept installing 0.1.0 even if I rm-rf node_modules

staltz:
I can update to 0.2.1 in a child PR

staltz (approved):

PR #154, opened 2021-04-19
Preprocess the conversion of strings to buffers for slowEquals and use LRU cache

Barbarrosa:
This MR includes only the performance enhancement, but you can reference https://github.com/Barbarrosa/jitdb/pull/18 which tests this using the (not yet merged) `slowEquals` benchmark.

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -1,5 +1,4 @@
 const bipf = require('bipf')
-const LRU = require('lru-cache')
 const traverse = require('traverse')
 const promisify = require('promisify-4loc')
 const pull = require('pull-stream')
@@ -32,12 +31,10 @@ function toBufferOrFalsy(value) {
   return Buffer.isBuffer(value) ? value : Buffer.from(value)
 }
 
-const slowEqualsCache = new LRU({
-  max: 1500,
-})
+const seekFromDescCache = new Map()
 function seekFromDesc(desc) {
-  if (slowEqualsCache.has(desc)) {
-    return slowEqualsCache.get(desc)
+  if (seekFromDescCache.has(desc)) {
+    return seekFromDescCache.get(desc)
   }
   const keys = desc.split('.').map(Buffer.from)
   // The 2nd arg `start` is to support plucks too
@@ -49,7 +46,7 @@ function seekFromDesc(desc) {
     }
     return p
   }
-  slowEqualsCache.set(desc, fn)
+  seekFromDescCache.set(desc, fn)
   return fn
 }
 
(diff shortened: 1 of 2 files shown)


operators.js
```diff
@@ -31,17 +32,25 @@ function toBufferOrFalsy(value) {
   return Buffer.isBuffer(value) ? value : Buffer.from(value)
 }
 
+const slowEqualsCache = new LRU({
+  max: 1500,
+})
```
staltz, line 37: Max 1500 is a lot, I suspect that we will never pass that threshold, making this LRU effectively the same as a (ES6) `Map`. Can I suggest that you change this `LRU` for a `Map`? That would also help with the bundle size to not include the lru-cache library. Also, the name of this cache could be `seekFromDescCache` because this is used also for `slowIncludes`, not just `slowEquals`.

Otherwise, this PR seems like a nice and simple improvement to perf.

Barbarrosa: @staltz Renamed & switched to `Map`. I favored LRU to support use cases with lots of dynamic keys, but perhaps that can wait until someone needs to address such cases.

Barbarrosa:
I reran the benchmarks, and this still performs better than the original `slowEquals`. These benchmarks include the CRC changes.

# [Original](https://github.com/Barbarrosa/jitdb/pull/16#issuecomment-822967343)

Part | Speed | Heap Change | Samples
-- | -- | -- | --
Query 1 big index with slowEqual (1st run) | 1610.01ms ± 7.02ms | -46.35 kB ± 0 kB | 33
Query 1 big index with slowEqual (2nd run) | 313.02ms ± 4.39ms | 26.37 kB ± 0 kB | 38
Query 1 big index with slowEqual (3rd run) | 0.34ms ± 0.03ms | 9.21 kB ± 0 kB | 40

# [Pre-calculate buffers w/ Map](https://github.com/Barbarrosa/jitdb/pull/22#issuecomment-822953322)

Part | Speed | Heap Change | Samples
-- | -- | -- | --
Query 1 big index with slowEqual (1st run) | 1395.08ms ± 15.25ms | -53.17 kB ± 0 kB | 38
Query 1 big index with slowEqual (2nd run) | 303.44ms ± 2.47ms | 75.09 kB ± 0 kB | 44
Query 1 big index with slowEqual (3rd run) | 0.45ms ± 0.17ms | 9.73 kB ± 34.88 kB | 42

staltz (approved):
Thanks! @arj03 What do you think?

arj03:
Sweet, thanks!

PR #155, opened 2021-04-19
Refactors related to Crc PR

staltz:


arj03:
`detectMissingAndLazyIndexes` might be refactored even more. Not sure if it will harder to read then.

staltz:
@arj03 How? What kind of changes are you imagining?

arj03:
Seems like the 2 functions now do almost the same thing, so could be refactored to patch in the check in the equals branch.

staltz:
What do you think of this one?

arj03 (approved):
Beautiful. Thank you

PR #156, opened 2021-04-24
Benchmark bugfixes

Barbarrosa:
I noticed several issues with my prior benchmark changes.
- The ability to run the full benchmark beginning-to-end with a single command was broken.
- The right side of the heap result was based on standard deviation, not margin of error.

This PR fixes the first issue and switches the right side of the heap result to use a calculation method more similar to [what nodemark itself uses](https://github.com/JoshuaWise/nodemark/blob/master/lib/result.js#L14), i.e. based on margin of error.


arj03:
Thanks

PR #157, opened 2021-04-24
Change "if" to "else if" to prevent unnecessary index generation

Barbarrosa:
Based on the documentation, it seems like the call here to `updatePrefixIndex` shouldn't occur when `useMap` is set.

> An additional option `useMap` can be provided that will store the prefix as a map instead of an array. The map can be seen as an inverted index that allows for faster queries at the cost of extra space. Maps don't store empty values meaning they are also a good fit for sparce indexes such as vote links.

### DIFF
--- a/index.js
+++ b/index.js
@@ -610,7 +610,7 @@ module.exports = function (log, indexesPath) {
               seq,
               offset
             )
-          if (op.data.prefix)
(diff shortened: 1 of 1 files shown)


arj03:
That is correct. Nice find!

PR #165, opened 2021-07-29
fix Buffer.from usage in operators.js

staltz:
Just stumbled upon this dangerous use of `Buffer.from`. 

Basically `arr.map(Buffer.from)` is equivalent to `arr.map((str, index, array) => Buffer.from(str, index, array))` which is not what we intended to do. Gladly, calling Buffer.from with those 3 args **still** yields the correct result. But we're just lucky. This PR changes the usage so that it obviously does not pass the 3 args.


arj03 (approved):
Right, good catch

PR #166, opened 2021-08-11
support live GT / GTE / LT / LTE operations

staltz:
1st :x: , 2nd :heavy_check_mark:

### DIFF
--- a/index.js
+++ b/index.js
@@ -235,6 +235,24 @@ module.exports = function (log, indexesPath) {
     }
   }
 
+  function seekMinTimestamp(buffer) {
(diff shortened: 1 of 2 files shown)


staltz:
I tried this out in ssb-index-feed-writer and it fixes the tests for live index-writing too. :+1:

arj03 (approved):
Great! I was a bit confused about the test until I saw the 2 keys :) Must be a bit groggy. Anyway I remember @happy0 at one point was asking about jitdb and live queries, so maybe he accidentally ran into this problem also.

Issue #163, opened 2021-07-21
deferred() to support cancellation


staltz:
## Problem

(Read a verbose context below, here is just the summary:) if there is a long-running async job running in a `deferred()` operator, there is no way of cancelling that job even if we have given up on the query. This is because deferred is based on simple callbacks and promises.

## Solution A

One way to fix this, which is a breaking change (anyway using `deferred` would have to update their code): change `deferred()` to be based on pull-streams, or any other async primitive that supports cancellations (I believe: pull-stream, push-stream, Callbags, RxJS, etc). This affects both the internals of `executeDeferredOps` as well as how the `deferred`s are used in practice by app developers.

## Solution B

Another way, less disruptive, but more ad-hoc: change the arguments of `deferred()`, from `deferred((meta, cb) =>` to `deferred((meta, cb, onAbort) =>`. This would avoid a breaking change, and `onAbort` is used like this:

```js
  deferred((meta, cb, onAbort) => {
    let timer = setTimeout(() => {
      timer = null
      cb(null, slowEqual('value.author', alice.id))
    }, 30e3)
    
    onAbort(() => {
      if (timer) {
        clearTimeout(timer)
        timer = null
      }
    })
  })
```

## Context

In a full-text search plugin (custom ssb-db2 index), suppose the user is going to type `little red riding hood`, but they first type `little`, then `red`, then `riding`, then `hood`, and suppose the app developer is triggering a search query every time a word is completed (as opposed to only triggering it once upon pressing keyboard Enter).

We will trigger 4 different searches: 

- `little` 
- `little red`
- `little red riding`
- `little red riding hood`

How can we cancel the previous searches? Once `little red` starts running, we should cancel the indexing reading task for the obsolete `little` search. Doing this with pull-streams is rather easy, because if there was a source pull-stream for `little`, we just abort that stream and start pulling the next one.

But I'd like to provide my plugin as an ssb-db2 operator, similar to fullMentions, so you can compose the search results with other ssb-db2 operations. The basic use case for this will be just to search for *public* `little red riding hood` messages of type `post`.

But the way that fullMentions does it is through `deferred` and a single callback, inside that `deferred` we are doing a `pull-level` read of the index, and `pull.collect` of all the results and then returning that to the callback. And this way of doing stuff doesn't allow cancellation.

staltz:
@arj03 I think I like solution B, what are your thoughts on that?

arj03:
[...]

PR #168, opened 2021-08-17
deferred() to support cancellation

staltz:
For issue #163 

This gave me a bit of headache to write, so I don't expect it to be a simple PR to review, but it seems correct and I'm glad our test suite is thorough.

The basic idea is to replace Promises in `executeDeferredOps` with pull-streams and `multicb`, because pull-streams support cancellation and it seems like a natural choice of async mechanism for our repo. The pull-stream `readable` replaces the `async-await` and `multicb` replaces the `Promise.all`.

When the readable is aborted (called with 1st arg `true`), then we call all the abort listeners.

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -2,7 +2,7 @@ const bipf = require('bipf')
 const traverse = require('traverse')
 const promisify = require('promisify-4loc')
 const pull = require('pull-stream')
-const pullAsync = require('pull-async')
+const multicb = require('multicb')
 const pullAwaitable = require('pull-awaitable')
 const cat = require('pull-cat')
 const { safeFilename } = require('./files')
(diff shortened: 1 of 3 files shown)


staltz:
@arj03 Any chance you could look at this one too? :)

arj03:
Right, thanks for the ping. I'll look at it this afternoon.

operators.js
```diff
@@ -330,56 +330,86 @@ function asOffsets() {
...
 
-  // Replace all deferreds with their respective results
-  allResults.forEach(([path, result]) => {
-    result.meta = meta
-    if (path.length === 0) ops = result
-    else traverse.set(ops, path, result)
-  })
+  return function readable(end, cb) {
+    if (end) {
+      while (abortListeners.length) abortListeners.shift()()
```
arj03, line 359: I remember I once ran into a performance problem with this pattern but I think it should be okay here because one has to explicitly add an abort handler to get this triggered.

arj03 (approved):
I looked over the changes, checked and ran the tests, looked over the new code in my editor. Can't see anything wrong with this new code. Again very happy we have all the existing tests. Nice one!

Issue #167, opened 2021-08-11
Reconsider pagination API


staltz:
Just a mild idea, not something I intend to code on immediately, but I don't want the idea to get lost in my brain:

From https://github.com/ssb-ngi-pointer/ssb-index-feed-writer/pull/5 I realized that maybe we should embed `pull.map(pull.values)` + `pull.flatten()` in JITDB when pagination is used, it's quite a common pattern and may feel like a gotcha. I bet there may be people who may want to get the page **array** and there will be people who just want the performance of pagination but in reality want a pull-stream of individual messages.

PR #170, opened 2021-08-17
add operator batch()

staltz:
For issue #167 

Achieved in a non-breaking manner, `paginate()` remains as it is, causing the final pull-stream to emit arrays of messages. We introduce `batch()` that has the same performance profile as `paginate()`, but just configure how the results should be streamed: one message at a time, no array of messages.

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -323,6 +323,10 @@ function paginate(pageSize) {
   return (ops) => updateMeta(ops, 'pageSize', pageSize)
 }
 
+function batch(batchSize) {
+  return (ops) => updateMeta(ops, 'batchSize', batchSize)
+}
+
 function asOffsets() {
   return (ops) => updateMeta(ops, 'asOffsets', true)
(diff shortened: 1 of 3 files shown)


operators.js
```diff
@@ -396,9 +400,9 @@ function toPullStream() {
     function paginateStream(ops) {
       let seq = meta.seq || 0
       let total = Infinity
-      const limit = meta.pageSize || 1
+      const limit = meta.pageSize || meta.batchSize || 20
```
staltz, line 403: The default `1` was actually pretty bad for performance. Reminder: https://github.com/ssb-ngi-pointer/jitdb/pull/123#issuecomment-782734363

arj03 (approved):
Thanks, this looks good and is a quite small change in the number of lines. 

Very happy we are integrating these best patterns into the library. Feels like next level in a way from fixing bugs :)

PR #172, opened 2021-08-23
reset any index if its version was bumped

staltz:
:x: :heavy_check_mark:

### DIFF
--- a/files.js
+++ b/files.js
@@ -46,6 +46,10 @@ function saveTypedArrayFile(filename, version, offset, count, tarr, cb) {
       if (err) console.error(err)
     }
 
+  if (typeof version !== 'number') {
+    return cb(new Error('cannot save file ' + filename + ' without version'))
+  }
+
   const dataBuffer = toBuffer(tarr)
   // we try to save an extra 10% so we don't have to immediately grow
   // after loading and adding again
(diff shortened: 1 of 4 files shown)


arj03:
So you are moving version around so that whenever you put something in indexes[] it must have a version, right? I thought I found a bug related to lazy loaded indexes but I can see that once it is loaded the version will be assigned. I'm really afraid that we make a mistake somewhere and forget this version number. There are only really 2 places ([1](https://github.com/ssb-ngi-pointer/jitdb/blob/64a67028618453cc5c2f912a320ff1a6a7e40107/files.js#L54), [2](https://github.com/ssb-ngi-pointer/jitdb/blob/64a67028618453cc5c2f912a320ff1a6a7e40107/files.js#L92)) in files.js where this is used when saving. What do you think about adding a `|| 1` there as a safeguard?

staltz:
Yeah I can add the safeguard. The reason why I removed it was actually to surface additional bugs in those cases where `index.version` does not exist, just to make sure that we have `index.version` defined in all cases.

Alternatively, I can also do something like throw an error in files.js in case version is not defined.

arj03:
Throw is probably better

staltz:
Updated!

arj03 (approved):
Assuming the tests pass :)

PR #174, opened 2021-09-13
add operator predicate()

staltz:
## Context

For ssb-meta-feeds, we're going to need to scan through all bendy butt messages and build an in-memory Map of subfeed IDs to their creation details (feedpurpose etc).

In order to scan through all bendy butt messages we need to do something like this

```js
ssb.db.query(
  where(valueAuthorIsBendyButtURI),
  toCallback(cb)
)
```

## Problem

I realize that we can't use `equal` because there are several bendybutt author URIs, not just one, and we can't use `includes` (because it's for arrays), we can't use what JITDB gives us. I would have to make a leveldb index, but it would still be a "bit vector"ish index, because all I need to know is a boolean, whether the given msg is authored by any bendybutt author URI or not.

## Solution

`predicate()` is a new operator in JITDB that allows you to pass it a boolean-returning function, similar to array.filter, e.g.

```js
slowPredicate('value.author', SSBURI.isBendyButtV1FeedSSBURI)
```

`predicate()` was pretty simple to implement because it's just a generalization of `equal()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,12 +103,15 @@ desired set of messages: `and`, `or`, `not`, `equal`, `slowEqual`, and others.
   - `objPath` a string in the shape `"foo.bar.baz"` which specifies the nested field `"baz"` inside `"bar"` inside `"foo"`
   - `value` is the same as `value` in the `equal` operator
   - `opts` same as the opts for `equal()`
-- `includes(seek, value, opts)` filters for messages where a `seek`ed` _field_
+- `includes(seek, value, opts)` filters for messages where a `seek`ed _field_
   is an array and includes a specific _value_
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -102,6 +103,14 @@ desired set of messages: `and`, `or`, `not`, `equal`, `slowEqual`, and others.
   - `objPath` a string in the shape `"foo.bar.baz"` which specifies the nested field `"baz"` inside `"bar"` inside `"foo"`
   - `value` is the same as `value` in the `equal` operator
   - `opts` same as the opts for `equal()`
+- `includes(seek, value, opts)` filters for messages where a `seek`ed` _field_
```
arj03, line 106: great you added these as well, can see they were missing

arj03 (approved):
Looks correct. Great addition!

PR #177, opened 2021-10-19
Reindex

arj03:
This allows jitdb to be able to reindex messages. The most common use cases for this is that some encrypted messages can now be decrypted because you got the decryption key.

~I'm marking this as a draft because I want to make sure that the [db2 changes](https://github.com/ssb-ngi-pointer/ssb-db2/pull/253/files) are fully ready~

In the end, all of this functionality is tested in the https://github.com/ssb-ngi-pointer/ssb-db2-box2 repo.

### DIFF
--- a/test/reindex.js
+++ b/test/reindex.js
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2021 Anders Rune Jensen
+//
+// SPDX-License-Identifier: Unlicense
+
 const test = require('tape')
 const validate = require('ssb-validate')
 const ssbKeys = require('ssb-keys')


index.js
```diff
@@ -1460,13 +1460,74 @@ module.exports = function (log, indexesPath) {
     )
   }
 
+  function reindex(offset, cb) {
+    let seq = 0
+    if (offset === 0) {
+      seq = 0
+      offset = -1
+    } else if (offset !== -1) {
```
staltz, line 1468: Would be good to have a code comment describing what this block is trying to achieve. So far I'm 2min looking at it trying to understand it.

index.js
```diff
@@ -1460,13 +1460,74 @@ module.exports = function (log, indexesPath) {
...
+      const { tarr } = indexes['seq']
+      let prevOffset = 0
+      for (const len = tarr.length; seq < len; ++seq)
```
staltz, line 1471: I feel a little unsafe with the for-loop not having tuborgs. Usually for-loops with no tuborgs have just one line under them, but this has 4. If someone wants to add one more line to this, it could become a footgun.

index.js
```diff
@@ -1460,13 +1460,74 @@ module.exports = function (log, indexesPath) {
...
+        if (tarr[seq] === offset) {
+          offset = prevOffset
```
staltz, line 1473: Perhaps it would also help for code readability that the function argument `offset` is not mutated. This code block is reading from `offset` but also reading from `offset`. It sort of changes the semantics of the variable. As a function argument, this is the `firstInvalidOffset`, while the calculation is discovering the `lastValidOffset`. Maybe you can find better names, but these are quick suggestions.

index.js
```diff
@@ -1460,13 +1460,74 @@ module.exports = function (log, indexesPath) {
...
+          }
+        }
+
+        index.offset = offset
+      }
+    }
+
+    push(
+      push.values(Object.entries(indexes)),
+      push.asyncMap(([indexName, index], cb) => {
```
staltz, line 1495: Neat use of `values(Object.entries` and `asyncMap`. Oh, and it's push-stream. I have a vague memory that push-stream `asyncMap` was incomplete or unmaintained. Tread carefully

staltz:
Overall very neat that you found a way to implement this as just one function. I put some comments about code readability and minor fixes.

test/reindex.js
```diff
@@ -0,0 +1,195 @@
...
+      return originalPipe(o)
+    }
+    return stream
+  }
+
+  function removeFilter() {
+    raf.stream = originalStream
+  }
+
+  return removeFilter
```
staltz, line 48: Nice!

staltz (approved):

PR #180, opened 2021-12-03
Fix null & undefined

arj03:
First commit should fail. Second fixes.

This is a PR to fix various equal bugs. I'll do a separate PR for each.

### DIFF
--- a/index.js
+++ b/index.js
@@ -297,9 +297,17 @@ module.exports = function (log, indexesPath) {
     }
   }
 
+  const nullBipf = bipf.allocAndEncode(null)
+  const undefinedBipf = bipf.allocAndEncode(undefined)
+
   function checkEqual(opData, buffer) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -297,9 +297,17 @@ module.exports = function (log, indexesPath) {
...
   }
 
+  const nullBipf = bipf.allocAndEncode(null)
+  const undefinedBipf = bipf.allocAndEncode(undefined)
+
   function checkEqual(opData, buffer) {
     const fieldStart = opData.seek(buffer)
-    if (!opData.value) return fieldStart === -1
+
+    if (opData.value === undefined && fieldStart === -1) return true
```
arj03, line 306: undefined is fine for missing fields, not is null

staltz (approved):
Looks correct!

PR #181, opened 2021-12-03
Add test for equal true/false

arj03:
First test a failure. This is actually the one that fails migrations in ssb-db2.

### DIFF
--- a/index.js
+++ b/index.js
@@ -303,8 +303,8 @@ module.exports = function (log, indexesPath) {
   function checkEqual(opData, buffer) {
     const fieldStart = opData.seek(buffer)
 
-    if (fieldStart === -1) return opData.value.equals(undefinedBuffer)
     // works for undefined
+    if (fieldStart === -1) return opData.value.equals(undefinedBuffer)
     else if (opData.value.length === 0)
       return bipf.compare(buffer, fieldStart, nullBipf, 0) === 0
     else return bipf.compareValue(buffer, fieldStart, opData.value, 0) === 0
--- a/index.js
+++ b/index.js
@@ -297,31 +297,17 @@ module.exports = function (log, indexesPath) {
     }
   }
 
+  const undefinedBuffer = bipf.slice(bipf.allocAndEncode(undefined), 0)
   const nullBipf = bipf.allocAndEncode(null)
-  const undefinedBipf = bipf.allocAndEncode(undefined)
-  const B_TRUE = Buffer.alloc(1, 1)
-  const trueBipf = bipf.allocAndEncode(true)
-  const B_FALSE = Buffer.alloc(1, 0)
-  const falseBipf = bipf.allocAndEncode(false)
 
   function checkEqual(opData, buffer) {
     const fieldStart = opData.seek(buffer)
 
-    if (opData.value === undefined && fieldStart === -1) return true
-    else if (opData.value === undefined)
-      return bipf.compare(buffer, fieldStart, undefinedBipf, 0) === 0
-    else if (opData.value === null)
+    if (fieldStart === -1) return opData.value.equals(undefinedBuffer)
+    // works for undefined
+    else if (opData.value.length === 0)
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -308,6 +312,10 @@ module.exports = function (log, indexesPath) {
       return bipf.compare(buffer, fieldStart, undefinedBipf, 0) === 0
     else if (opData.value === null)
       return bipf.compare(buffer, fieldStart, nullBipf, 0) === 0
+    else if (opData.value.equals(B_TRUE) || opData.value === true)
```
arj03, line 315: note we use the buffer version inside isPrivate()

staltz: Are these for backwards compatibility?
I was expected that `opData.value` would be equal to `trueBipf`. Or at least we could make it clear that **all** opData.value are BIPF **or** all opData.value are plain JavaScript values. This `if` condition mixes things up.

staltz: Basically I would be more comfortable with `opData.value` being strictly always a BIPF buffer, because then we can just use `bipf.compare()` directly. Notice also that this function `checkEqual` is gradually becoming a reimplementation of `bipf.encode` since we take JS values in the `if` condition and match them with the BIPF values in the `return`.

arj03: hmm yeah, I'll see if there is better way to do this

staltz: I'm smelling that there may be a breaking change to `jitdb` here, but I think we can do that and still not have a breaking change in `ssb-db2`. Maybe it will cause reindexing to happen, but maybe we find ways to avoid the reindex.

arj03: The existing behavior was that `value` would be either string (auto converted to buffer) or a buffer encoding the value. Undefined/null was used for the field not being defined. https://github.com/ssb-ngi-pointer/jitdb/pull/180 changed the rules a bit and made special cases for null & undefined so they would be auto converted similar to string. Poking a bit more at this issue I see that it would have been better for them to have been converted inside `toBufferOrFalsy` in operators.js because checkEqual is a hot path so we want that doing minimal work. 

Before we were relying on compareString working a bit like a compareValue where you would just check the actual buffer part because of the bug in bipf. I was trying out changing equal to take a bipf encoded value, but that is a larger and breaking change. And it also touches things like prefixes. So instead I was wondering what the minimal change could be and tried out a compareValue function instead. I'm a bit unsure what the best solution is. I pushed up a commit that does the minimal changes and uses https://github.com/ssbc/bipf/pull/23.

staltz: I think `compareValue` is a decent idea because things were working before with `compareString` (albeit a buggy/fragile situation), so the new method would be conservative and most likely work in JITDB. The thing we need to be careful though is to make sure this won't cause bugs in the future. It would be good to get it right sooner, rather than later. One way it could go wrong (I'm not sure if this is would happen) is if the "value" part of the bipf buffer is not enough information, like two "values" are equal but the "type" tag is different. For example `<Buffer 61 62>` encoded to BIPF is `<Buffer 11 61 62>`, and `"ab"` encoded to BIPF is `<Buffer 10 61 62>`. The "value" part is the same, and could pass the equality check, but it would be an error because the decoded values are different.

Also, what if `undefined` ends up being encoded in BIPF and stored in the log as `0e 02`, and then what if people want to query for that? I think it would be safer against bugs that `opData.value` is *always* the same type, and not to have special logic depending on what it is. So, dropping strings and dropping undefined. We could have a special `opData` for testing absence.

Mostly I'm concerned that we should get this right now that we have spotted that this problem.

arj03: One thing a change to bipf values means is that it will force us to be strict about the values we send in. So no more: https://github.com/ssb-ngi-pointer/ssb-db2/blob/39e2563199bca6f4895d5970135bf4db2f122048/operators/index.js#L163 instead we must use true which is a good thing, but does is a breaking change. I'll try and do patches for jitdb and ssb-db2 in a separate PR and then we can compare.

Issue #178, opened 2021-11-18
Use bipf.compare() instead of bipf.compareString()


staltz:
Full context here: https://github.com/ssb-ngi-pointer/ssb-db2/pull/277

BIPF was recently fixed because it had some wrong logic, but then the updated-fixed BIPF broke tests in ssb-db2. That's because we are using `bipf.compareString()` in JITDB for comparing things that are **not strings**, e.g. `kvt.meta.private`.

We should update BIPF in JITDB and replace `bipf.compareString()` with `bipf.compare()` wherever applicable. We should also be careful to do `bipf.allocAndEncode()` instead of `Buffer.from()` for the opData.value.

arj03:
I'll take a stab at this one :)

PR #182, opened 2021-12-06
Change the value of an operation to be a bipf value

arj03:
This is a breaking change. Luckily it is only breaking for clients calling with an operation directly instead of using `equal` and friends. This PR aims to solve #178 by changing values to always be bipf encoded instead of buffers. This change also makes things a bit clearer on how equal works I think. Either you give a specific value in which case the value inside the buffer must match exactly (this includes null, undefined, booleans and also numbers that was not working properly before) OR the field is not defined in which cases undefined also matches.

This PR grew out of https://github.com/ssb-ngi-pointer/jitdb/pull/181 which is an attempt to solve the same problem in a backwards compatible way. The changes needed to ssb-db2 to work with this branch is surprisingly few because most thing rely on equal, that automatically handles this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -297,23 +297,13 @@ module.exports = function (log, indexesPath) {
     }
   }
 
-  const nullBipf = bipf.allocAndEncode(null)
   const undefinedBipf = bipf.allocAndEncode(undefined)
 
   function checkEqual(opData, buffer) {
     const fieldStart = opData.seek(buffer)
 
-    if (opData.value === undefined && fieldStart === -1) return true
-    else if (opData.value === undefined)
-      return bipf.compare(buffer, fieldStart, undefinedBipf, 0) === 0
-    else if (opData.value === null)
-      return bipf.compare(buffer, fieldStart, nullBipf, 0) === 0
-    else if (
-      ~fieldStart &&
-      bipf.compareString(buffer, fieldStart, opData.value) === 0
-    )
-      return true
-    else return false
+    if (fieldStart === -1 && opData.value.equals(undefinedBipf)) return true
+    else return bipf.compare(buffer, fieldStart, opData.value, 0) === 0
   }
 
   function compareWithRangeOp(op, value) {
@@ -354,23 +344,20 @@ module.exports = function (log, indexesPath) {
     const fieldStart = opData.seek(buffer)
     if (!~fieldStart) return false
     const type = bipf.getEncodedType(buffer, fieldStart)
-    if (type === bipf.types.string) {
-      return checkEqual(opData, buffer)
-    } else if (type === bipf.types.array) {
+
+    if (type === bipf.types.array) {
       let found = false
       bipf.iterate(buffer, fieldStart, (_, itemStart) => {
         const valueStart = opData.pluck
           ? opData.pluck(buffer, itemStart)
           : itemStart
-        if (bipf.compareString(buffer, valueStart, opData.value) === 0) {
+        if (bipf.compare(buffer, valueStart, opData.value, 0) === 0) {
           found = true
           return true // abort the bipf.iterate
         }
       })
       return found
-    } else {
-      return false
-    }
+    } else return checkEqual(opData, buffer)
   }
 
   function safeReadUint32(buf, prefixOffset = 0) {
@@ -843,7 +830,7 @@ module.exports = function (log, indexesPath) {
   function matchAgainstPrefix(op, prefixIndex, cb) {
     const target = op.data.value
     const targetPrefix = target
-      ? safeReadUint32(target, op.data.prefixOffset)
+      ? safeReadUint32(bipf.slice(target, 0), op.data.prefixOffset)
       : 0
     const bitset = new TypedFastBitSet()
     const bitsetFilters = new Map()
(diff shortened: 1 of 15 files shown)


arj03:
I ran the benchmarks locally and I get almost the exact same numbers as master with this branch.

operators.js
```diff
@@ -101,8 +96,8 @@ function debug() {
 function slowEqual(seekDesc, target, opts) {
   opts = opts || {}
   const seek = seekFromDesc(seekDesc)
-  const value = toBufferOrFalsy(target)
-  const valueName = !value ? '' : value.toString()
+  const value = bipf.allocAndEncode(target)
+  const valueName = !target ? '' : `${target}`
```
arj03, line 100: We need to be really careful with these so that the indexes has the exact same names as before. The tests in ssb-db2 actually failed without this change :)

index.js
```diff
@@ -297,23 +297,13 @@ module.exports = function (log, indexesPath) {
...
-    else if (opData.value === null)
-      return bipf.compare(buffer, fieldStart, nullBipf, 0) === 0
-    else if (
-      ~fieldStart &&
-      bipf.compareString(buffer, fieldStart, opData.value) === 0
-    )
-      return true
-    else return false
+    if (fieldStart === -1 && opData.value.equals(undefinedBipf)) return true
+    else return bipf.compare(buffer, fieldStart, opData.value, 0) === 0
```
staltz, line 306: :+1:

index.js
```diff
@@ -354,23 +344,20 @@ module.exports = function (log, indexesPath) {
     const fieldStart = opData.seek(buffer)
     if (!~fieldStart) return false
     const type = bipf.getEncodedType(buffer, fieldStart)
-    if (type === bipf.types.string) {
-      return checkEqual(opData, buffer)
-    } else if (type === bipf.types.array) {
+
+    if (type === bipf.types.array) {
```
staltz, line 348: :+1:

index.js
```diff
@@ -843,7 +830,7 @@ module.exports = function (log, indexesPath) {
   function matchAgainstPrefix(op, prefixIndex, cb) {
     const target = op.data.value
     const targetPrefix = target
-      ? safeReadUint32(target, op.data.prefixOffset)
+      ? safeReadUint32(bipf.slice(target, 0), op.data.prefixOffset)
```
staltz, line 833: What's the story behind this?

arj03: for prefix we must match the actual buffer and not the type + buffer. bipf.slice will give us the buffer so that it works like before where target was the buffer.

index.js
```diff
@@ -297,23 +297,13 @@ module.exports = function (log, indexesPath) {
...
-      return bipf.compare(buffer, fieldStart, undefinedBipf, 0) === 0
-    else if (opData.value === null)
-      return bipf.compare(buffer, fieldStart, nullBipf, 0) === 0
-    else if (
-      ~fieldStart &&
-      bipf.compareString(buffer, fieldStart, opData.value) === 0
-    )
-      return true
-    else return false
+    if (fieldStart === -1 && opData.value.equals(undefinedBipf)) return true
```
staltz, line 305: I'm a bit afraid of this, because BIPF (unlike JSON) will encode undefined as actual fields, and maybe one day we might want to `checkEqual` for those undefined fields, but this logic would prevent looking for those fields.

```js
BIPF.allocAndEncode({a: undefined, b: 3})
/*
<Buffer 5d 08 61 0e 02 08 62 22 03 00 00 00>
              |  _____    |     ___________
              a    |      b          |
                 undefined           3
              
*/
```

```
JSON.stringify({a: undefined, b: 3})
// '{"b":3}'
```

How about instead we have a special operator `absent()` which has `op.type == "ABSENT"` and then we implement a `checkAbsent()` function similar to `checkEqual()`?

index.js
staltz: Oh :+1: , TIL how bipf.slice actually works

index.js
arj03: It should work in both cases. If you have something that is json encoded and you put that into bipf, like your example then we will hit this case: `fieldStart === -1 && opData.value.equals(undefinedBipf)` and it will work. On the other hand if you encoded bipf directly, then `a` will be there and be equal to `undefined`. In this case the second bipf.compare should work. So I don't see the need for a new operator?

staltz: If `fieldStart >= 0` and the opData.value matches undefinedBipf, then it will use `bipf.compare`, yeah.

But I mean that from the perspective of the developer's intention. When they write `equal(seekRoot, undefined, {`, what does the developer want? If they want to find `root` fields that are exactly `undefined`, but instead they get messages where the `root` field doesn't exist, then they are getting results that are not what they were looking for.

staltz: This is hypothetical, but I'm thinking that it's better to be safe and not creating a gotcha.

arj03: I think it would be a bigger gotcha with another type. This smells a bit like [3 valued logic](https://en.wikipedia.org/wiki/Three-valued_logic) in SQL where there is this indeterminate third value. As a developer writing something like the isRoot equal function I would expect that `undefined` would just work and I didn't have to think about it. Because I would be putting in a undefined value into the database, so I would expect that I can then query for undefined to get it out again.

staltz: Okay, we got on a call and agreed to do both things: keep the `undefined` logic as is in this PR, plus introduce `absent()` operator.

staltz:
I also updated bipf to 1.5.4.

After CI is green, let's merge?

PR #183, opened 2021-12-08
add absent() and slowAbsent() operators

staltz:


operators.js
```diff
@@ -41,7 +41,7 @@ function seekFromDesc(desc) {
     var p = start
     for (let key of keys) {
       p = bipf.seekKey(buffer, p, key)
-      if (!~p) return void 0
+      if (p < 0) return -1
```
staltz, line 44: I changed this because then this function consistently returns numbers and allows us to do `p < 0` comparisons. `undefined < 0` doesn't work because it returns false.

arj03 (approved):
Looks good. Thanks :)

Issue #185, opened 2021-12-20
Problematic writes


arj03:
Saving a core index is done [async](https://github.com/ssb-ngi-pointer/jitdb/blob/41a1b326a682b8c9a18f9b7dd7283021d31cc428/index.js#L176). This means we could end up in a situation where an index is written before things like `seq.index`, which could lead to [bad things](https://github.com/ssb-ngi-pointer/jitdb/blob/41a1b326a682b8c9a18f9b7dd7283021d31cc428/index.js#L478) if it crashes or was forced to terminate in that particular point in time.

staltz:
True! Good that you found this, I had no suspicions about that part. Do you have a fix in mind already? Like, we could add callbacks on the saving core indexes, and when they are done, we save the non-core index. But we also have to be careful with the Nth saving of index A to wait for the `N-1`th saving of A.

staltz:
If `seq.index` is broken, could that affect the leveldb bug directly? Because `keys` index is based on the `seq()` operator.

For instance:

Suppose `type_post.index` was saved but `seq.index` wasn't saved, as you originally described.

Like (each uppercase letters is a record):

```
      seq.index: A B C D E F G
type_post.index: A B C D E F G H I J K
```

Then, supposed JITDB could update the `type_post.index` in [update index](https://github.com/ssb-ngi-pointer/jitdb/blob/41a1b326a682b8c9a18f9b7dd7283021d31cc428/index.js#L517).

The [startSeq](https://github.com/ssb-ngi-pointer/jitdb/blob/41a1b326a682b8c9a18f9b7dd7283021d31cc428/index.js#L493) would be **H** (because the `for` loop would never find the common `offset` value between `seq.index` and `type_post.index` and would just run through the whole for loop), but the `gt` to [start the stream](https://github.com/ssb-ngi-pointer/jitdb/blob/41a1b326a682b8c9a18f9b7dd7283021d31cc428/index.js#L517) would be **K**, because `gt` is just the offset of the `type_post.index`.

Then, when the stream write happens, we get record **L** and we [update seq index](https://github.com/ssb-ngi-pointer/jitdb/blob/41a1b326a682b8c9a18f9b7dd7283021d31cc428/index.js#L523) to so that **H's seq would point to L's offset**. That's clearly the wrong offset.

[...]

PR #186, opened 2021-12-21
Wait for core indexes to be saved before writing other indexes

arj03:
This is to fix https://github.com/ssb-ngi-pointer/jitdb/issues/185

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,6 @@ const TypedFastBitSet = require('typedfastbitset')
 const bsb = require('binary-search-bounds')
 const multicb = require('multicb')
 const FastPriorityQueue = require('fastpriorityqueue')
-const promisify = require('promisify-4loc')
 const debug = require('debug')('jitdb')
 const debugQuery = debug.extend('query')
 const Status = require('./status')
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -497,18 +498,24 @@ module.exports = function (log, indexesPath) {
     const indexNeedsUpdate = !coreIndexNames.includes(op.data.indexName)
 
     function save(count, offset) {
-      if (updatedSeqIndex) saveCoreIndex('seq', indexes['seq'], count)
+      const coreWrites = []
```
staltz, line 501: How about multicb instead? In this file we already use multicb (and that's because it was faster than Promise.all) and there are so far zero promises in this file.

staltz: *I can't promise* (pun intended)

arj03: Good call :)

index.js
```diff
@@ -497,18 +498,23 @@ module.exports = function (log, indexesPath) {
...
       if (updatedSequenceIndex)
-        saveCoreIndex('sequence', indexes['sequence'], count)
+        saveCoreIndex('sequence', indexes['sequence'], count, done())
 
       index.offset = offset
       if (op.data.version > index.version) index.version = op.data.version
 
-      if (indexNeedsUpdate) saveIndex(op.data.indexName, index, count)
+      if (indexNeedsUpdate) {
+        done(() => {
```
staltz, line 514: Just a doubt: is it possible that none of the core indexes would need updates? And what happens with the done then?

arj03: That works, have tested :)

staltz (approved):
Ideally there would be a test for this, but I won't insist on it at all, I know how hard it must be to reproduce the bug in tests, and it's christmas. Time for some green button pressing

arj03:
True, happy holidays

Issue #190, opened 2022-01-13
`absent` operator -- an example of `seek`


nichoth:
This is part of this sbot -- https://github.com/planetary-social/planetary-pub/blob/7ea9eb2fd3b021f1f862eaea9b0349a5fb0c2d97/viewer.js#L133

What is the type of value does `absent` take?

```
absent('value.root', { indexType: 'roots' }),
```

```
/Users/nick/code/planetary-pub/node_modules/jitdb/index.js:345
    const fieldStart = opData.seek(buffer)
                              ^

TypeError: opData.seek is not a function
    at checkAbsent (/Users/nick/code/planetary-pub/node_modules/jitdb/index.js:345:31)
    at updateIndexValue (/Users/nick/code/planetary-pub/node_modules/jitdb/index.js:427:38)
    at /Users/nick/code/planetary-pub/node_modules/jitdb/index.js:687:16
    at Array.forEach (<anonymous>)
    at Object.write (/Users/nick/code/planetary-pub/node_modules/jitdb/index.js:668:23)
    at Object.o.write (/Users/nick/code/planetary-pub/node_modules/ssb-db2/log.js:95:30)
    at Stream._writeToSink (/Users/nick/code/planetary-pub/node_modules/async-append-only-log/stream.js:60:33)
    at Stream._handleBlock (/Users/nick/code/planetary-pub/node_modules/async-append-only-log/stream.js:96:12)
    at Stream._resumeCallback (/Users/nick/code/planetary-pub/node_modules/async-append-only-log/stream.js:143:24)
    at Request._callback (/Users/nick/code/planetary-pub/node_modules/async-append-only-log/index.js:151:9)
```

arj03:
See the tests :) 

https://github.com/ssb-ngi-pointer/jitdb/blob/5d197adfe551888a8ba813a1ab84394e12c18130/test/operators.js#L1475

It needs to be a [function](https://github.com/ssb-ngi-pointer/jitdb/blob/5d197adfe551888a8ba813a1ab84394e12c18130/test/helpers.js#L93). This should be in the README.

staltz:
Actually, absent() is explained in the README already. See previous definitions of the `seek` argument in the README, a bit above the description of absent()

nichoth:
So, looking at #191 -- it's not clear to me what these example values are used for. And what does the `seekChannel` function do here?

```js
const B_VALUE = Buffer.from('value')
[...]

PR #191, opened 2022-01-14
add 'absent' example to readme

nichoth:
This adds an example to the readme

The example was taken directly from the test file, and shows the `seek` function as well.

Any help with the way things are worded would be appreciated

related to #190

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ const bipf = require('bipf')
 const assert = require("assert");
 const B_VALUE = Buffer.from('value')
 const B_CHANNEL = Buffer.from('channel')
+const B_CONTENT = Buffer.from('content')
 
 query(
   fromDB(db),
--- a/README.md
+++ b/README.md
@@ -232,6 +232,7 @@ query(
   ),
   toCallback((err, msgs) => {
     assert.equal(err, null, 'got no error')
+    // do something with msgs...
   })
 )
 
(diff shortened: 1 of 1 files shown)


arj03:
Thanks. We already have an example with seekType that explains how these functions can be constructed. It seems to me that what was confusion was maybe the signature of `seek`? If that is the case, then maybe we can call it `seekFn` instead?

nichoth:
That's interesting I think maybe I totally missed the `seekType` example in the readme; I had looked at the tests in this repo. One thing I would add though is where the various things are required:

```js
const { query, fromDB, where, and, toCallback } = require('jitdb/operators')
const bipf = require('bipf')
```

From the [readme](https://github.com/ssb-ngi-pointer/jitdb#operators) v

```js
query(
  fromDB(db),
  where(equal(seekType, 'post', { indexType: 'type' })),
  toCallback((err, msgs) => {
    console.log('There are ' + msgs.length + ' messages of type "post"')
  })
)

// The `seekType` function takes a buffer and uses `bipf` APIs to search for
// the fields we want.
const bValue = Buffer.from('value') // better for performance if defined outside
const bContent = Buffer.from('content')
const bType = Buffer.from('type')
function seekType(buffer) {
  var p = 0 // p stands for "position" in the buffer, offset from start
  p = bipf.seekKey(buffer, p, bValue)
  if (p < 0) return
  p = bipf.seekKey(buffer, p, bContent)
  if (p < 0) return
  return bipf.seekKey(buffer, p, bType)
}
```

PR #193, opened 2022-03-14
Sort by

arj03:
# Background

I was working on adding a [list method](https://gitlab.com/ahau/lib/ssb-crut/-/merge_requests/25) to ssb-crut together with @mixmix and we got to talk about sorting of the results. We would like to support to modes of sorting: creation time and receive (or arrival) time. Right now results from jitdb are sorted by creation time, unless creation time is after arrival time, in which case we use arrival time. This is mostly to handle computers with wrong local time writing messages in the *future*. This can then be returned as ascending or descending. The purpose for this is sorting data where data from one or more peers are delayed. Lets say you have a list of latest comments or something like that. And you sort by creation time. Later you receive some messages from a peer that was offline the first time around. The results might then be displayed in the middle of the list instead of top. While both are equally true, depending on how this is displayed in the UI it might be good to support both methods of ordering.

What this PR does is add a `sortBy` operator that defaults to existing creation time behaviour. If set to `arrival` the results will be sorted by when they are written to disc instead of creation time.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1241,11 +1241,22 @@ module.exports = function (log, indexesPath) {
     limit,
     descending,
     onlyOffset,
+    sortBy,
     cb
   ) {
     seq = seq || 0
 
-    const sorted = sortedByTimestamp(bitset, descending)
+    let sorted
+    if (sortBy === 'arrival') {
+      sorted = new FastPriorityQueue()
+      // already sorted by arrival time in log
+      sorted.heapify(
+        bitset.array().map((x) => {
+          return { seq: x }
+        })
+      )
+    } else sorted = sortedByTimestamp(bitset, descending)
+
     const resultSize = sorted.size
 
     // seq -> record buffer
@@ -1313,7 +1324,12 @@ module.exports = function (log, indexesPath) {
     else return bitset.size() - seq
   }
 
-  function paginate(operation, seq, limit, descending, onlyOffset, cb) {
+  function paginate(operation, seq, limit, descending, onlyOffset, sortBy, cb) {
+    if (!cb) {
+      cb = sortBy
+      sortBy = 'declared'
+    }
+
     onReady(() => {
       const start = Date.now()
       executeOperation(operation, (err0, result) => {
@@ -1326,6 +1342,7 @@ module.exports = function (log, indexesPath) {
           limit,
           descending,
           onlyOffset,
+          sortBy,
           (err1, answer) => {
             if (err1) cb(err1)
             else {
@@ -1346,7 +1363,12 @@ module.exports = function (log, indexesPath) {
     })
   }
 
-  function all(operation, seq, descending, onlyOffset, cb) {
+  function all(operation, seq, descending, onlyOffset, sortBy, cb) {
+    if (!cb) {
+      cb = sortBy
+      sortBy = 'declared'
+    }
+
     onReady(() => {
       const start = Date.now()
       executeOperation(operation, (err0, result) => {
@@ -1359,6 +1381,7 @@ module.exports = function (log, indexesPath) {
           Infinity,
           descending,
           onlyOffset,
+          sortBy,
           (err1, answer) => {
             if (err1) cb(err1)
             else {
--- a/index.js
+++ b/index.js
@@ -1197,13 +1197,14 @@ module.exports = function (log, indexesPath) {
     const fpq = new FastPriorityQueue(
       descending ? compareDescending : compareAscending
     )
-    bitset.forEach((seq) => {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1182,20 +1189,37 @@ module.exports = function (log, indexesPath) {
...
+
+  function compareSeqDescending(a, b) {
+    return a.seq > b.seq
+  }
+
+  function sortedBy(bitset, descending, sortBy) {
     updateCacheWithLog()
     const order = descending ? 'descending' : 'ascending'
+    const sortedCache = sortBy === 'arrival' ? sortedSeqCache : sortedTSCache
+    const comparer =
```
arj03, line 1204: The linter butchered this long line. It was originally like this:

```
    const comparer = sortBy === 'arrival' ?
          descending ? compareSeqDescending : compareSeqAscending :
          descending ? compareDescending : compareAscending
```

index.js
```diff
@@ -1182,20 +1189,37 @@ module.exports = function (log, indexesPath) {
...
     if (sortedCache[order].has(bitset)) return sortedCache[order].get(bitset)
-    const fpq = new FastPriorityQueue(
-      descending ? compareDescending : compareAscending
-    )
-    bitset.forEach((seq) => {
-      fpq.add({
-        seq,
-        timestamp: indexes['timestamp'].tarr[seq],
+    const fpq = new FastPriorityQueue(comparer)
+    fpq.heapify(
```
arj03, line 1215: If you look at the commits I refactored this. I noticed there was this heapify method that does exactly what we want so it should be faster but I couldn't detect this when put into ssb-db2 as there is a bit too much noise in results (+/- 5% up and down).

index.js
```diff
@@ -1313,7 +1338,12 @@ module.exports = function (log, indexesPath) {
     else return bitset.size() - seq
   }
 
-  function paginate(operation, seq, limit, descending, onlyOffset, cb) {
+  function paginate(operation, seq, limit, descending, onlyOffset, sortBy, cb) {
+    if (!cb) {
```
arj03, line 1342: I ended up defaulting this so we don't have to break the API of `paginate` and `all`. Can change if needed.

arj03:
I plugged this into the link PR and the tests are now running for db2 the same way they do for db1.

index.js
staltz: It's fine, I know the feeling, let's let the linter own styling on our behalf

index.js
```diff
@@ -1182,20 +1189,37 @@ module.exports = function (log, indexesPath) {
...
     const order = descending ? 'descending' : 'ascending'
+    const sortedCache = sortBy === 'arrival' ? sortedSeqCache : sortedTSCache
+    const comparer =
+      sortBy === 'arrival'
+        ? descending
+          ? compareSeqDescending
+          : compareSeqAscending
+        : descending
+        ? compareDescending
+        : compareAscending
```
staltz, line 1211: Could `compareDescending`/`Ascending` be renamed to `compareTSDescending`/`Ascending` so that it's consistent with the naming of `compareSeqDescending`?

index.js
staltz: Sweet

operators.js
```diff
@@ -387,6 +387,10 @@ function descending() {
   return (ops) => updateMeta(ops, 'descending', true)
 }
 
+function sortBy(sortOption) {
+  return (ops) => updateMeta(ops, 'sortBy', sortOption)
+}
+
```
staltz, line 393: Instead of this, could we have the operators `sortByArrival` and `sortByCreation`? It's easy to make mistakes with string APIs, and those mistakes would be totally silent (no error thrown). This would make it similar to `descending()`, note that we don't have the operator `order('descending')` and `order('ascending')`, we only have `descending()`. In that light, we can also drop `sortByCreation()` and have only `sortByArrival()`, since the default is creation.

index.js
staltz: I would tend towards a breaking change, and bumping ssb-db2, it doesn't take that much work. But maintaining this legacy mode of either 6 or 7 arguments seems like something we could burn ourselves with in the future.

staltz:
@arj03 Overall looks like a pretty simple change, thanks! I had just a few minor comments and then we can merge.

arj03:
Thanks. Pushed up some review fixes.

index.js
staltz: @arj03 How about this?

arj03: ohh noo. I didn't see this. I'll do a new PR with a breaking change and publish that as 5.0?

PR #195, opened 2022-03-14
update async-append-only-log to 4.0.0

staltz:


arj03 (approved):
Looks good. Is the peer dependency needed?

PR #198, opened 2022-03-22
Implement group by

arj03:
# Background

@mixmix was working on adding [an option](https://gitlab.com/ahau/lib/ssb-crut/-/merge_requests/28) to crut list where you would get results ordered by updateTime. We would like to avoid running through all the results (db1 has to do this) in order to display say the latest 5 buttcasts. For this what we really need is a group by operator similar to what you have in a normal database. So this is quite general, you could also use this functionality to show the latest post for each feed. I tried modelling this similar to the seek function. This means we still have to run through results to extract what we need, but since we are working directly on the bipf data, it's not really that bad. This can still be combined with regular where filtering and when used with limit, it will stop once enough results are found.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1267,12 +1267,47 @@ module.exports = function (log, indexesPath) {
     descending,
     onlyOffset,
     sortBy,
+    groupBy,
     cb
   ) {
     seq = seq || 0
 
-    const sorted = sortedBy(bitset, descending, sortBy)
-    const resultSize = sorted.size
+    let sortedBitset = sortedBy(bitset, descending, sortBy)
+
+    if (groupBy) {
+      let seqs = sortedBitset.kSmallest(Infinity).map((x) => x.seq)
+      const uniqueByValue = new Map()
+      push(
+        push.values(seqs),
+        push.asyncMap(getRecord),
+        push.drain(
+          (record) => {
+            const fieldStart = groupBy(record.value)
+            if (fieldStart < 0) return true
+            const value = bipf.decode(record.value, fieldStart)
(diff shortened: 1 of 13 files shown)


index.js
```diff
@@ -1267,12 +1267,47 @@ module.exports = function (log, indexesPath) {
...
+            const fieldStart = groupBy(record.value)
+            if (fieldStart < 0) return true
+            const value = bipf.decode(record.value, fieldStart)
+            if (!uniqueByValue.has(value))
+              uniqueByValue.set(value, record.value)
+
+            if (uniqueByValue.size == seq + limit) return false
+          },
+          (early) => {
+            if (early) return
```
arj03, line 1294: not sure why we need this with push.drain

mixmix:
I don't quite understand the need for this but trust you. 

to be clear, the way I get the 5 most recently updated eg podcasts is pull for type podcast (getting updates and roots). then run this through crut.read and throw out the duds. once I have 5 good ones kill the stream.

oh, and we have a pull.unique in there too.
is this what groupBy would be helping with.?

arj03:
> to be clear, the way I get the 5 most recently updated eg podcasts is pull for type podcast (getting updates and roots). then run this through crut.read and throw out the duds. once I have 5 good ones kill the stream.
> 
> oh, and we have a pull.unique in there too. is this what groupBy would be helping with.?

It is the same algorithm, now that you explain it. So right now this is probably only marginally faster. It does maybe improve the discoverability of these kind of things. 

I do have an idea to use an index for this, then it should be faster. So challegence accepted :-) Changing this to draft.

Issue #199, opened 2022-03-25
Reindex after or during log compaction


staltz:
async-append-only-log now supports `log.compact()` https://github.com/ssb-ngi-pointer/async-append-only-log/issues/48 and we should reindex the relevant portions of jitdb indexes.

staltz:
@arj03 I'm working on this already, but I bumped into a design issue. My original plan is to rebuild all indexes while log compaction is still ongoing, based on what [Dominic suggested here](https://github.com/ssb-ngi-pointer/ssb-db2/issues/306#issuecomment-1034479270)

> Instead of writing a separate second log, superimpose the compacted log on top of the old one, keeping track of the min uncompacted sequence number. When you do a query from the old indexes, drop results with offsets smaller than the uncompacted sequence, and mix with results from the new indexes. merge index responses by sequence numbers, so numbers smaller than the uncompacted sequence are from the compacted section and the numbers greater are from the uncompacted log.

For instance, my plan is to create a folder `.ssb/db2/indexes/_compacting` where the new index files exist, and once those are done catching up with the log, we delete all jitdb indexes in `.ssb/db2/indexes` and move up the files in `.ssb/db2/indexes/_compacting` to `.ssb/db2/indexes`.

However, does this go against the JIT design of JITDB?

Alternatively, should we just delete all JITDB indexes once log compaction is done, and let the JIT nature take care of rebuilding them?

arj03:
Will get back to you on this tomorrow :)

staltz:
@arj03 Thanks!

`reindex()` currently doesn't rebuild the core indexes (seq, sequence, timestamp), is there a reason for that? I think with compaction we would also have to rebuild the core indexes.

Why do you say that level indexes should be allowed to lookup while jitdb shouldn't? Is it because of speed of rebuilding the indexes? (By the way, in my tests, I felt that some jitdb indexes do take a considerable time to build, like `value_author.32prefix`, which is 6 MB in a production database, and has to write 4 bytes for each record.)

arj03:
> reindex() currently doesn't rebuild the core indexes (seq, sequence, timestamp), is there a reason for that? I think with compaction we would also have to rebuild the core indexes.

Right, that is because for encrypted they don't change. Here they do :)

> Why do you say that level indexes should be allowed to lookup while jitdb shouldn't?

Just because it is easier. We don't need to flag to jitdb that says: please don't update indexes, just run on what you have.

[...]

PR #200, opened 2022-03-29
use binary search to get seq from offset

staltz:
## Context

I'm preparing for #199 and one thing I realized is that I'll need a quick way of getting the seq from any given offset, e.g. during compaction there is going to be `unshiftedOffset` and I'd like to discover `unshiftedSeq`.

## Problem

We don't have a helper function for that and currently we're just doing a linear for-loop to look for the offset.

Another problem is that the "end" of the for-loop is `tarr.length`, which I think is wrong because the tarr grows much more than the `count` does (like it doubles in size every time it "grows" but most of it is still unused). Instead we need to use `index.count`.

## Solution

Introduce `getSeqFromOffset` that does a binary search and uses `indexes['seq'].count` as the end marker.

This should also have a performance improvement for those cases when the jitdb indexes are many and large but we just need to update the tip of the index. This way we can avoid scanning the whole index in `O(n)` and can do `O(log n)`. I'm not sure we have any benchmark for this but should be a pretty common thing to happen in production.

### DIFF
--- a/index.js
+++ b/index.js
@@ -841,8 +841,8 @@ module.exports = function (log, indexesPath) {
     const seqs = []
     opOffsets.sort((x, y) => x - y)
     const opOffsetsLen = opOffsets.length
-    const { tarr } = indexes['seq']
-    for (let seq = 0, len = tarr.length; seq < len; ++seq) {
+    const { tarr, count } = indexes['seq']
+    for (let seq = 0; seq < count; ++seq) {
       if (bsb.eq(opOffsets, tarr[seq]) !== -1) seqs.push(seq)
       if (seqs.length === opOffsetsLen) break
     }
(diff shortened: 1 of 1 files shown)


staltz:
Here's previous benchmarks to compare with: https://github.com/ssb-ngi-pointer/jitdb/pull/197

But there's no noticeable difference. But do we have a benchmark to measure this case? (A big index receives a bit of updates)

arj03:
This looks like a great optimization. I'm wondering if we can't we use [bsb](https://www.npmjs.com/package/binary-search-bounds) that is already included?

staltz:
Good point! I switched to using bsb.eq in getSeqFromOffset, and I took a look at the source code, it's basically the same, so that's good.

I also added another commit [use seq index.count instead of index.tarr.length](https://github.com/ssb-ngi-pointer/jitdb/pull/200/commits/bcba09eff01981fa8e384883adfafcdbbbd6580f).

arj03 (approved):
Excellent. This is a good improvement.

PR #201, opened 2022-03-29
dont prune old fields from status object

staltz:
## Context

Some Manyverse users have been complaining that the progress bar for "preparing database" should never decrease, it should always monotonically increase.

(By the way, this PR is part of my Manyverse work, not part of the log compaction grant)

## Problem

The `pruneStatus()` functionality in `./status.js` was originally intended to align with the "JIT" nature of JITDB, but in reality it doesn't help at all, because it means that certain fields sometimes disappear (due to `pruneStatus()`) and then appear (due to the index being updated), making the overall progress bar calculation change a lot.

## Solution

The "JIT" nature should only be per **session**, that is, the indexes that are loaded into memory in the current session should have, respectively, their `status` fields always available.

I removed `pruneStatus()` and updated tests accordingly.

### DIFF
--- a/status.js
+++ b/status.js
@@ -10,19 +10,9 @@ module.exports = function Status() {
   const obv = Obv()
   obv.set(indexesStatus)
   const EMIT_INTERVAL = 1000 // ms
-  const PRUNE_INTERVAL = 2000 // ms
   let i = 0
   let iTimer = 0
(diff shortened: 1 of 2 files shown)


arj03 (approved):
👍

PR #202, opened 2022-03-29
add prepare() API to bypass laziness

staltz:
## Context

Some indexes in Manyverse are always required but the index only begins building after leveldb indexes (which are eager) are built. This may cause some sudden jumps backwards in the progress bar.

## Problem

For indexes that we know are always required, we need to have a way of turning off laziness, so to trigger their building as soon as possible. This also helps to have these indexes registered in `status`.

## Solution

New API `prepare()` which "warms up" the index. See the readme diff for a description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,6 +567,13 @@ will be the messages directly.
 
 Similar to `all` except it does not fetch records from the log, it only responds with a number that tells the total amount of records matching the operation.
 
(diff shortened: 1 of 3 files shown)


arj03 (approved):
Yeah makes sense, should be faster than doing a `all` query on it.

PR #203, opened 2022-03-30
prepare() updates status when starting

staltz:

### DIFF
--- a/test/add.js
+++ b/test/add.js
@@ -428,6 +428,8 @@ prepareAndRunTest('indexAll multiple reindexes', dir, (t, db, raf) => {
 })
 
 prepareAndRunTest('prepare an index', dir, (t, db, raf) => {
+  t.plan(10)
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1033,8 +1033,8 @@ module.exports = function (log, indexesPath) {
     } else console.error('Unknown type', op)
   }
 
-  function traverseEqualsAndIncludes(operation, fn) {
-    function traverseMore(ops) {
+  function forEachLeafOperationIn(operation, fn) {
```
staltz, line 1036: I renamed this because nowadays we support EQUAL, INCLUDES, PREDICATE, ABSENT, so the full name would be `traverseEqualsAndIncludesAndPredicateAndAbsent` which is pretty horrible, so I went with this one instead.

arj03: :)

arj03 (approved):
Looks good 👍

PR #205, opened 2022-03-31
Improve status calculation and teardown

staltz:
As you may have guessed, I worked on this all day. (Context https://github.com/ssb-ngi-pointer/jitdb/pull/204 and https://github.com/ssb-ngi-pointer/ssb-db2/pull/322)

I tried all sorts of things, and I was coding directly in the Manyverse node_modules, so the approach I have in this PR has already been tested in Manyverse, both with ssb-fixtures and with real production data.

The summary is:

`status.js` (not `index.js`) needs to keep track of the indexes being updated, and once **all** of _those_ are done, then reset the status. The concept of a "progress bar session" exists in `status.js`, not in `index.js`. 

This was tricky because sometimes `index.js` "ended" processing all indexes, only to immediately start processing _another_ set of indexes, so we couldn't rely on that "end" event.

Additionally, I removed the status update from `executeOperation` because that was the wrong spot: `paginate()` calls `executeOperation` many times even when the indexes are already built, so we don't want to trigger status for those cases. So I moved it to the beginning of prepare().

### DIFF
--- a/index.js
+++ b/index.js
@@ -70,7 +70,7 @@ module.exports = function (log, indexesPath) {
       }
     }
 
-    status.batchUpdate(indexes, coreIndexNames)
+    status.update(indexes, coreIndexNames)
 
     isReady = true
     for (let i = 0; i < waiting.length; ++i) waiting[i]()
@@ -525,6 +525,7 @@ module.exports = function (log, indexesPath) {
 
     const logstreamId = Math.ceil(Math.random() * 1000)
     debug(`log.stream #${logstreamId} started, to update index ${waitingKey}`)
+    status.update(indexes, indexNamesForStatus)
 
     log.stream({ gt: index.offset }).pipe({
       paused: false,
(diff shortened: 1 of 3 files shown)


arj03:
This looks correct, one thing I was wondering is that once these are done, you clear the indexes, so does that mean that the overall progress in db2 will go backwards, or how do you keep track of that?

staltz:
@arj03 Ah yes, that was one of the dilemmas I ran into. I solved it with [this commit in the ssb-db2 branch](https://github.com/ssb-ngi-pointer/ssb-db2/pull/322/commits/826bd42588847316178fcf83256caac1d7b84e0d) and the gist of it is to calculate the total progress like this:

```
average_jitdb_offset = (average(progress_for_each_jitdb_index) || 1) * logSize
total = average(all_leveldb_offsets + average_jitdb_offset) / (leveldb_count + 1)
```

Basically, once the jitdb index statuses are "cleared", we consider the its progress to be `1` (100%) and thus `average_jitdb_offset` is equal the logSize.

arj03 (approved):
👍

staltz:
Oh, and even though this PR removes `resetStatus` as a breaking change, I'm not gonna make a major version because almost surely no one else used it.

PR #206, opened 2022-04-02
Improve performance of query when seq is far behind log

arj03:
I noticed this when doing a query: first jitdb updates seq and only after that, it updates what is needed for the query. There is an assumption that seq is in sync with the log after running a query, but instead of eagerly updated only `seq` index and then the rest. Why not try updating the indexes we need and only after that `seq` if needed. In most cases the `seq` update will then be a no op.

Running my normal 1GB query on empty db this is 15.7s compared to 17.6s before. I know this is a bit of a worst case example and in most real cases the difference won't be very big as `seq` is often very close to being in sync with the log because it is updated on any query.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,8 +57,8 @@
     "ssb-keys": "^8.1.0",
     "ssb-ref": "^2.14.3",
     "ssb-validate": "^4.1.1",
+    "tap-arc": "^0.3.4",
     "tap-bail": "^1.0.0",
-    "tap-arc": "^0.1.2",
     "tape": "^5.2.2"
   },
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -57,8 +57,8 @@
     "ssb-keys": "^8.1.0",
     "ssb-ref": "^2.14.3",
     "ssb-validate": "^4.1.1",
+    "tap-arc": "^0.3.4",
```
arj03, line 60: I was battling a bit with the Update index test that uses the `status` call and updated this while tested. Locally that test seems to randomly fail, might be related to our test runner, but still it shouldn't matter in what order the tests are run.

arj03:
This now also fixes #207

staltz (approved):
Seems very correct! Thanks

PR #208, opened 2022-04-03
pValue

arj03:
I was wondering if we could do better on initial sync. #206 started out as that. That PR changed 1 initial jitdb query from around 17s to 15 something. Then I started wondering if we could do more and found that doing these extracting values (seq, timestamp, sequence) was significant. Meaning not only what we are looking for (opData.seek), but also the base values. Then I started playing around and found that actually finding the position of value inside the buffer is quite significant, so I refactored the code to get that once and pass that in. Note seekFromDesc allows a start (defaults to 0) and because of that we need 3 parameters to the seek function. I changed the seekers in ssb-db2 and now things are down to around 11s. This is on my laptop when it is not plugged in. Plugged in we are down to **7s** :-)

This PR is against #206, I'll change that once we know about that PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -257,32 +257,28 @@ module.exports = function (log, indexesPath) {
     }
   }
 
-  function seekMinTimestamp(buffer) {
+  function seekMinTimestamp(buffer, pValue) {
     var p = 0 // note you pass in p!
     p = bipf.seekKey(buffer, p, B_TIMESTAMP)
     const arrivalTimestamp = bipf.decode(buffer, p)
-    p = 0
-    p = bipf.seekKey(buffer, p, B_VALUE)
-    p = bipf.seekKey(buffer, p, B_TIMESTAMP)
+    p = bipf.seekKey(buffer, pValue, B_TIMESTAMP)
     const declaredTimestamp = bipf.decode(buffer, p)
     return Math.min(arrivalTimestamp, declaredTimestamp)
   }
 
-  function seekSequence(buffer) {
-    var p = 0 // note you pass in p!
-    p = bipf.seekKey(buffer, p, B_VALUE)
-    p = bipf.seekKey(buffer, p, B_SEQUENCE)
+  function seekSequence(buffer, pValue) {
+    var p = bipf.seekKey(buffer, pValue, B_SEQUENCE)
     return bipf.decode(buffer, p)
   }
 
-  function updateTimestampIndex(seq, offset, buffer) {
+  function updateTimestampIndex(seq, offset, buffer, pValue) {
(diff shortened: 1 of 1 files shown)


staltz:
Very good idea! I'll review properly tomorrow, but the idea is 💯

staltz:
What about also pContent? It should be quite common to go into msg.value.content

arj03:
It really depends on how often it is used. pValue is used in both timestamp, sequence and in most user specified queries. If you did a query for only author, then pContent would be wasted. pContent might make more sense in level indexes inside db2. I tried added it, but I don't have enough plugins to see any difference. In manyverse I have a feeling you are using a lot so might make more sense.

index.js
```diff
@@ -315,8 +311,8 @@ module.exports = function (log, indexesPath) {
 
   const undefinedBipf = bipf.allocAndEncode(undefined)
 
-  function checkEqual(opData, buffer) {
-    const fieldStart = opData.seek(buffer)
+  function checkEqual(opData, buffer, pValue) {
+    const fieldStart = opData.seek(buffer, 0, pValue)
```
staltz, line 315: In what implementations of `opData.seek` will we need the 2nd argument `0`? I took a look at the ssb-db2 PR, and it seems we're never using this 2nd arg

arj03: It is used here: https://github.com/ssb-ngi-pointer/jitdb/blob/6d62f3acc6935c6a900ed07a8bb805ca6504281b/operators.js#L40. At first I just had pValue as the first argument after buffer, but that broke includes + pluck test ;)

staltz, line 315: In what implementations of `opData.seek` will we need the 2nd argument `0`? I took a look at the ssb-db2 PR, and it seems we're never using this 2nd arg

index.js
```diff
@@ -334,11 +330,14 @@ module.exports = function (log, indexesPath) {
   }
 
   function checkComparison(op, buffer) {
+    let pValue = 0
+    pValue = bipf.seekKey(buffer, pValue, B_VALUE)
```
staltz, line 334:
```js
const pValue = bipf.seekKey(buffer, 0, B_VALUE)
```

I don't see any upside in the `let` style

staltz (approved):
Exciting to squeeze more performance out of this!

arj03:
I guess this should be 7.0?

staltz:
I think there is no breaking change here. Correct me if I'm wrong.

The new argument is at the end, so any consumer of jitdb can be free to not use that argument and just recalculate pValue.

arj03:
Right, javascript will just throw away any extra arguments :)

PR #209, opened 2022-04-04
use bipf 1.6.0 seekKeyCached

staltz:
Context: https://github.com/ssb-ngi-pointer/jitdb/pull/208

Uses BIPF 1.6.0's `seekKeyCached` API instead of passing pValue around.

@arj03 How did you run those benchmarks in your PR? Could you try running them with this code? I know you did a good job in #208, but I think `seekKeyCached` is going to help us with `pValueContent` and potentially others.

### DIFF
--- a/test/prefix.js
+++ b/test/prefix.js
@@ -277,9 +277,8 @@ prepareAndRunTest('Prefix equal', dir, (t, db, raf) => {
       value: helpers.toBipf(
         '%wOtfXXopI3mTHL6F7Y3XXNtpxws9mQdaEocNJuKtAZo=.sha256'
       ),
-      indexType: 'vote',
-      indexName:
-        'value_content_vote_link_%wOtfXXopI3mTHL6F7Y3XXNtpxws9mQdaEocNJuKtAZo=.sha256',
+      indexType: 'value_content_vote_link',
+      indexName: 'value_content_vote_link',
       prefix: 32,
     },
   }
--- a/index.js
+++ b/index.js
@@ -257,28 +257,29 @@ module.exports = function (log, indexesPath) {
     }
   }
 
-  function seekMinTimestamp(buffer, pValue) {
-    var p = 0 // note you pass in p!
-    p = bipf.seekKey(buffer, p, B_TIMESTAMP)
-    const arrivalTimestamp = bipf.decode(buffer, p)
-    p = bipf.seekKey(buffer, pValue, B_TIMESTAMP)
-    const declaredTimestamp = bipf.decode(buffer, p)
+  function seekMinTimestamp(buffer) {
+    const pTimestamp = bipf.seekKey(buffer, 0, B_TIMESTAMP)
+    const arrivalTimestamp = bipf.decode(buffer, pTimestamp)
+    const pValue = bipf.seekKeyCached(buffer, 0, B_VALUE)
+    const pValueTimestamp = bipf.seekKey(buffer, pValue, B_TIMESTAMP)
+    const declaredTimestamp = bipf.decode(buffer, pValueTimestamp)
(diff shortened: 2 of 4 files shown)


test/prefix.js
```diff
@@ -277,9 +277,8 @@ prepareAndRunTest('Prefix equal', dir, (t, db, raf) => {
       value: helpers.toBipf(
         '%wOtfXXopI3mTHL6F7Y3XXNtpxws9mQdaEocNJuKtAZo=.sha256'
       ),
-      indexType: 'vote',
-      indexName:
-        'value_content_vote_link_%wOtfXXopI3mTHL6F7Y3XXNtpxws9mQdaEocNJuKtAZo=.sha256',
+      indexType: 'value_content_vote_link',
+      indexName: 'value_content_vote_link',
```
staltz, line 281: Unrelated to this PR, I just stumbled upon this and decided to fix it so that these two fields would match what we would get from `operators.js`

arj03:
Yeah this is great. Can you create a db2 branch that builds on top of this? That makes it easier for me to test.

staltz:
Working on it!

staltz:
@arj03 One problem is that we have to have global and unique buffers for B_VALUE and others, otherwise we can end up having two B_VALUEs which are in different memory addresses, and the WeakMap idea won't work on them.

E.g.

```js
const a = Buffer.from('value')
const b = Buffer.from('value')

a === b // false

const weakMap = new WeakMap()
weakMap.set(a, 3)
weakMap.has(b) // false
```

arj03:
Ugh, would it be possible to pass in value as a string. I think they match the same address or am I confusing javascript with .net ;-)?

staltz:
Yeah, that's what I'm thinking about. `bipf.seekKey(buffer, start, target)` supports both string `target` and buffer `target`, which means we need to support both also for `bipf.seekKeyCached`. And thus I could use strings to do the WeakMap look ups. Which means that if you pass a buffer `target` in `bipf.seekKeyCached(buffer,start,target)` you would lose some performance benefits, but at least it would remain functioning correctly.

arj03:
My benchmark script, it's a bit hacky but gets the job done :)

```
const SecretStack = require('secret-stack')
const caps = require('ssb-caps')
const path = require('path')
const ssbKeys = require('ssb-keys')
const { where, and, type, isRoot, hasRoot, author, paginate, descending,
        toPullStream, toCallback } = require('../operators')
const bipf = require('bipf')
const pull = require('pull-stream')

const dir = './perf-testing/ssb'
const keys = ssbKeys.loadOrCreateSync(path.join('/home/arj/.ssb', 'secret'))

const ssb = SecretStack({ appKey: caps.shs })
  .use(require('../'))
  .use(require('../compat/ebt')) // ebt db helpers
  .use(require('../full-mentions')) // include index
  .call(null, {
    keys,
    path: dir,
    db2: {
      startDecryptBox1: "2022-03-25",
    }
  })

console.log('Doing the query')
console.time('query')

ssb.db.getJITDB().all(
  author('@QlCTpvY7p9ty2yOFrv1WU1AE88aoQc4Y7wYal7PFc+w=.ed25519'),
  0, false, false, null, (err, results) => {
    console.timeEnd('query')
    console.log(results.length)
  }
)
```

and then I delete all indexes to test both level + jitdb and only:

```
rm perf-testing/ssb/db2/indexes/seq* perf-testing/ssb/db2/indexes/timestamp.index perf-testing/ssb/db2/indexes/value_author.32prefix
```

to test only jitdb :)

And ./perf-testing/ssb/log.bipf is just a copy from ~/.ssb

staltz:
Parking this as draft for now because this PR can only be merged if it necessarily makes performance better.

Issue #210, opened 2022-04-04
Seekers should not return undefined


staltz:
We typically return undefined if the field pointer was -1:

https://github.com/ssb-ngi-pointer/jitdb/blob/1dc806047567b1321fe52494f76b8cb01d846c4a/test/helpers.js#L41

This could be a problem if we later check the return value of these seeker functions, like here:

https://github.com/ssb-ngi-pointer/jitdb/blob/1dc806047567b1321fe52494f76b8cb01d846c4a/index.js#L315-L317

Shouldn't we always return `-1` instead of `undefined`? Could this have been a source of bugs?

Note:

```
> ~-1
0

> !~-1
true

> ~0
-1

> !~0
false

> ~undefined
-1

> !~undefined
false
```

cc @arj03

arj03:
Yep, that looks incorrect

PR #211, opened 2022-04-06
fix seekers in test/helpers

staltz:
For issue #210. Doesn't affect code in production because it's just `test/helpers.js` but in ssb-db2 we have real seekers used in production. PR for that coming later.


arj03 (approved):
Yep

PR #213, opened 2022-04-07
fix crash in getSeqFromOffset

### LINKED ISSUE #212: getSeqFromOffset not found (crash from Manyverse)
```
App: se.manyver 0.2204.5-beta-indie (125)
Device: samsung SM-G781W (arm64-v8a | armeabi-v7a | armeabi)
OS: Android 12 (SDK 31)
User comment: Crash on Load. started every load about 15 hrs prior to this report. installed via APK.

Stack trace from Node.js:
getSeqFromOffset(779403143) not found
Error: getSeqFromOffset(779403143) not found
    at getSeqFromOffset (/data/user/0/se.manyver/files/nodejs-project/index.js:17531:24)
    at updateIndex (/data/user/0/se.manyver/files/nodejs-project/index.js:17714:41)
    at ensureIndexSync (/data/user/0/se.manyver/files/nodejs-project/index.js:18023:7)
    at ensureSeqIndexSync (/data/user/0/se.manyver/files/nodejs-project/index.js:18031:5)
    at AsyncMapStream.fn (/data/user/0/se.manyver/files/nodejs-project/index.js:18315:34)
    at AsyncMapStream.BbJ+AHGSnoS13OpDSkWGHhM+CJATmn9px1+2E8kOdtg=.AsyncMapStream.write (/data/user/0/se.manyver/files/nodejs-project/index.js:21054:8)
    at ValueStream.resume (/data/user/0/se.manyver/files/nodejs-project/index.js:9149:15)
    at AsyncMapStream.ThroughStream.resume (/data/user/0/se.manyver/files/nodejs-project/index.js:94747:81)
    at AsyncMapStream.ThroughStream.resume (/data/user/0/se.manyver/files/nodejs-project/index.js:94747:81)
    at AsyncMapStream.ThroughStream.resume (/data/user/0/se.manyver/files/nodejs-project/index.js:94747:81)
```

This is the line where the error is thrown: https://github.com/ssb-ngi-pointer/jitdb/blob/b6715d461d5087b88d8dfed93481218224dce1dd/index.js#L308

staltz:
Fixes #212 

I am not sure what triggered the `throw` (in a way it would be good to preserve the `throw` so that we can detect and study these cases!), but I'm reverting to returning 0 instead, that's because the [previously implementation was choosing 0 as the seq in case the offset wasnt found](https://github.com/ssb-ngi-pointer/jitdb/commit/540d008dc41b2a4fcc103cb07fbe709548d7f835), and we should probably have preserved that behavior.

### DIFF
--- a/index.js
+++ b/index.js
@@ -305,7 +305,7 @@ module.exports = function (log, indexesPath) {
     if (offset === -1) return 0
     const { tarr, count } = indexes['seq']
     const seq = bsb.eq(tarr, offset, 0, count - 1)
-    if (seq < 0) throw new Error(`getSeqFromOffset(${offset}) not found`)
+    if (seq < 0) return 0
     return seq
   }
 


arj03 (approved):
Yeah, not good to crash the whole app with a throw. I wonder what good form of logging there could be for this. Guess it really depends on the app, so manyverse has its own way for reporting errors.

staltz:
Yeah overall I want to improve the logging aspect a lot, but it has to be done right (open source and opt-in) and that takes time, so for now I want to make this quick fix. Thanks for the review

Issue #215, opened 2022-04-13
get just one result


staltz:
ssb-db2 `getMsg` API could be optimized. Instead of using jitdb `all()`, we can just do:

- `msgId` input
- `keys` leveldb index gives us the `seq`
- We pass the `seq` to a new API in jitdb that gives the `offset`
- We look up the record from the log using `offset`

I bet this is going to be great for performance because maybe setting up bitvectors and scanning them and all of that stuff might be non-negligible once you have lots of `getMsg` going on.

staltz:
@arj03 Actually I realized that I wanted to ask what you think about **this** issue, not the other one. :sweat_smile:

arj03:
:D I was also a bit surprised by the other one. Anyway, I think this optimization could make sense, I'm just not sure how often `getMsg` is called.

staltz:
getMsg would be called for every delete, which I think when there's many deletes it would become a big deal. And then in ssb-threads, getMsg is used often to get the root msg of a thread, and it processes many threads.

PR #216, opened 2022-04-18
add lookup() API

staltz:
Context: issue https://github.com/ssb-ngi-pointer/jitdb/issues/215 and improving the performance of deletes in ssb-db2.

I already confirmed that using this API and changing the implementation of `ssb-db2`'s `del()` made deletes become more than 10x faster, so this API should be worth adding.

It basically gives you raw access to index entries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -574,6 +574,13 @@ updating them, if necessary. This is not a query, as it won't return any
 results. When done, the callback `cb` is just called with the "duration" of
 index preparation as the second argument.
 
+### lookup(operation, seq, cb)
+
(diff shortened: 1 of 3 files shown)


staltz:
arj03 reviewed this PR (as an IM app message) and approved while also commenting that the index should be checked if it exists before using it.

arj03:
Api and code looks good 👍

-------- Originalbesked --------
Til 18. apr. 2022 11.08, github-actions[bot] skrev:

staltz:
Whut

PR #217, opened 2022-04-20
add APIs necessary for log compaction algorithms

staltz:
Context: https://github.com/ssb-ngi-pointer/jitdb/issues/199

**Problem 1:** we need to know when is indexing ongoing and/or when is querying ongoing, because we don't want compact to happen in parallel with those.

**Problem 2:** if compaction is ongoing, we can't allow JITDB queries to start.

**Problem 3:** `reindex` has to rebuild also the core indexes.

**Solution:**

- Introduces the public APIs `indexingActive` and `queriesActive` which are Obz's that emit numbers of how many are active at the moment. 
- Monitor the status of compaction and guard the queries to wait for compaction to end first
- `reindex` rebuilds also the core indexes

These changes are already used and tested in https://github.com/ssb-ngi-pointer/ssb-db2/pull/339

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,6 +40,7 @@ module.exports = function (log, indexesPath) {
   const indexes = {}
   let isReady = false
   let waiting = []
+  const waitingCompaction = []
   const coreIndexNames = ['seq', 'timestamp', 'sequence']
 
   loadIndexes(() => {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -1605,9 +1648,11 @@ module.exports = function (log, indexesPath) {
     push(
       push.values(Object.entries(indexes)),
       push.asyncMap(([indexName, index], cb) => {
-        if (coreIndexNames.includes(indexName)) return cb()
-
-        if (index.lazy) {
+        if (coreIndexNames.includes(indexName)) {
```
arj03, line 1651: right, reindex can now be done when things are removed, so this changes the core indexes. Before when it was only encrypted, these core indexes does not change.

arj03 (approved):
This was really easy to review. Nice work.

PR #218, opened 2022-04-26
Seekkey2

arj03:
See https://github.com/ssbc/bipf/pull/33

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,9 +93,9 @@ module.exports = function (log, indexesPath) {
     }
   })
 
-  const B_TIMESTAMP = Buffer.from('timestamp')
-  const B_SEQUENCE = Buffer.from('sequence')
-  const B_VALUE = Buffer.from('value')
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -93,9 +93,9 @@ module.exports = function (log, indexesPath) {
     }
   })
 
-  const B_TIMESTAMP = Buffer.from('timestamp')
-  const B_SEQUENCE = Buffer.from('sequence')
-  const B_VALUE = Buffer.from('value')
+  const B_TIMESTAMP = bipf.allocAndEncode('timestamp')
```
staltz, line 96: We use the convention `B_FOO` for `Buffer.from('foo')` so perhaps we can name these `FOO_BIPF` instead

staltz (changes requested):

index.js
arj03: I went with BIPF_

staltz (approved):
🚀

PR #219, opened 2022-06-27
add test for loading corrupted header tarr file

staltz:
**Context:** https://github.com/ssbc/ssb-db2/issues/358

**Problem:** an error is *thrown* from `readFileAndCheckCRC()` when it should be in fact `cb(err)`'d.

**Solution:** catch all errors in that function and call cb(err).

1st :x: 2nd :heavy_check_mark: and then additional commits just add extra checks.

### DIFF
--- a/files.js
+++ b/files.js
@@ -34,11 +34,18 @@ function calculateCRCAndWriteFile(buf, filename, cb) {
 function readFileAndCheckCRC(filename, cb) {
   readFile(filename, (err, buf) => {
     if (err) return cb(err)
+    if (buf.length < 16) return cb(Error('file too short'))
 
-    const crcFile = buf.readUInt32LE(3 * FIELD_SIZE)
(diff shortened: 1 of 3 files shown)


files.js
```diff
@@ -34,11 +34,18 @@ function calculateCRCAndWriteFile(buf, filename, cb) {
 function readFileAndCheckCRC(filename, cb) {
   readFile(filename, (err, buf) => {
     if (err) return cb(err)
+    if (buf.length < 16) return cb(Error('file too short'))
```
arj03, line 37: Looks good. This function is used in all other load function so checking here is the right place 👍

arj03 (approved):
Great

staltz:
CI is angry at me, but so am I angry at it.

arj03:
Latest one looks good, more error checking 👍

PR #220, opened 2022-06-30
ensure only one log.scan at a time

staltz:
## Context

I was experimenting with Manyverse doing deleteFeed-then-compact-then-reindex, and it was odd that post-compaction reindexing lasted 6.5sec while *initial* indexing lasted 4.0sec. 

## Problem

I investigated deeper and the reason was that initial indexing triggers jitdb `createIndexes` which does *one* log scan for *multiple* indexes, but reindexing triggers jitdb `updateIndex` which does one log scan for *each* index.

## Solution

Unify `updateIndex` and `createIndexes` such that both creation and update are performed by the same log.stream. This additionally ensures that jitdb *never* does concurrent log.streams, at any given point in time there is only one jitdb log.stream ongoing.

PS: The git diff is pretty nasty to read, but in reality the new implementation of `updateIndexes()` follows closely the structure of the two old functions.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ const path = require('path')
 const bipf = require('bipf')
 const push = require('push-stream')
 const pull = require('pull-stream')
+const mutexify = require('mutexify')
 const toPull = require('push-stream-to-pull-stream')
 const pullAsync = require('pull-async')
 const TypedFastBitSet = require('typedfastbitset')
@@ -206,6 +207,7 @@ module.exports = function (log, indexesPath) {
   }
 
   function saveIndex(name, index, count, cb) {
+    debug('saving index: %s', name)
     if (index.prefix && index.map) savePrefixMapIndex(name, index, count, cb)
     else if (index.prefix) savePrefixIndex(name, index, count, cb)
     else saveBitsetIndex(name, index, cb)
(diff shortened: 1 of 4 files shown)


staltz:
Previous benchmark results, for comparison: https://github.com/ssbc/jitdb/pull/219

Spoiler: looks okay

index.js
```diff
@@ -1134,20 +1028,19 @@ module.exports = function (log, indexesPath) {
...
+        const ops = []
+        ops.push({ data: { indexName: 'seq' } }) // always ensure seq updated
+        forEachLeafOperationIn(operation, (op) => {
+          ops.push(op)
+        })
+        updateIndexes(ops, next)
       }),
 
-      // ensure that the seq->offset index is synchronized with the log
-      push.asyncMap((_, next) => ensureSeqIndexSync(next)),
```
arj03, line 1149: this ensureSeqIndexSync is now only used in 1 place. Can we in that place just call  ensureIndexSync({ data: { indexName: 'seq' } }, .... and then remove that function?

index.js
```diff
@@ -843,34 +757,30 @@ module.exports = function (log, indexesPath) {
   }
 
   function filterIndex(op, filterCheck, cb) {
-    ensureIndexSync(op, () => {
```
arj03, line 846: are all of these remove ensureIndexSync because executeOperation now makes sure that indexes are ready, so we don't need to check this everywhere?

index.js
staltz: I'm now on mobile, but yes I'll do that later

staltz: done

arj03 (approved):
:star:

PR #221, opened 2022-07-12
rewind paginate() seq after compaction

staltz:
## Context

Putting compaction in Manyverse, and testing it with real world data, I stumbled upon a problem.

Bipf.decode was failing in various ways because the bipf.decode was starting at the wrong byte offset. Digging deeper, it turned out that the reason is that there are some db queries with `toPullStream` where the pull-stream source is pulled asynchronously (either because of `too-hot` or because of any other async processing). So there was one "pull" of the `toPullStream` which happened *after* compaction had begun. We have a `waitingCompaction` mechanism in place, such that this "pull" waits for compaction.

## Problem

When the "pull" resumed after compaction, that meant a call to `paginate()`, **using the same seq** as it had before, which would end up being mapped to an outdated offset, which caused the bipf.decode crash.

## Solution

When a `paginate()` is **resumed** after a compaction, we should update the `seq` by rewinding it, equivalent to the number of holes found in the log during compaction. This is why we depend on PR https://github.com/ssbc/async-append-only-log/pull/80 being merged first.

### DIFF
--- a/operators.js
+++ b/operators.js
@@ -535,7 +535,7 @@ function toPullStream() {
               else if (answer.total === 0) cb(true)
               else {
                 total = answer.total
-                seq += limit
+                seq = answer.nextSeq
                 cb(null, answer.results)
               }
             }
--- a/test/compaction.js
+++ b/test/compaction.js
@@ -1,0 +1,111 @@
+// SPDX-FileCopyrightText: 2022 Andre 'Staltz' Medeiros <contact@staltz.com>
+//
+// SPDX-License-Identifier: CC0-1.0
+
+const validate = require('ssb-validate')
+const ssbKeys = require('ssb-keys')
+const path = require('path')
+const rimraf = require('rimraf')
+const mkdirp = require('mkdirp')
+const pify = require('util').promisify
+const { prepareAndRunTest, addMsgPromise, helpers } = require('./common')()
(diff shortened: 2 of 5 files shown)


test/compaction.js
```diff
@@ -0,0 +1,111 @@
...
+    toPullStream()
+  )
+
+  const msgs = await pify(source)(null)
+  t.equals(msgs.length, 3, '1st page has 3 messages')
+  t.equals(msgs[0].value.content.text, 'Testing 0', 'msg 0')
+  t.equals(msgs[1].value.content.text, 'Testing 3', 'msg 3')
+  t.equals(msgs[2].value.content.text, 'Testing 4', 'msg 4')
+
+  // FIXME: AAOL needs to set stats.done = false as soon as compact() is called
```
staltz, line 84: Reminder to self to do this.

operators.js
```diff
@@ -535,7 +535,7 @@ function toPullStream() {
               else if (answer.total === 0) cb(true)
               else {
                 total = answer.total
-                seq += limit
+                seq = answer.nextSeq
```
arj03, line 538: this looks like a general bug, right? In that case :+1: nice find :)

arj03 (approved):
👍

PR #222, opened 2022-07-13
refactor internal listFiles function

staltz:
I thought that maybe `index.js` shouldn't care about browser-versus-nodejs and files.js can handle the check.


arj03 (approved):
Yep, makes sense

Issue #223, opened 2022-07-13
bitsets might need to be aware of holes


staltz:
I stumbled upon a big headache kind of problem. Pagination seems to give wrong results if there are deleted records (and so far this has nothing to do with compaction, you can assume we're not running any compaction).

Example below:

```js
  const content0 = { type: 'post', text: 'Testing 0' }
  const content1 = { type: 'post', text: 'Testing 1' }
  const content2 = { type: 'post', text: 'Testing 2' }
  const content3 = { type: 'post', text: 'Testing 3' }
  const content4 = { type: 'post', text: 'Testing 4' }
  const content5 = { type: 'post', text: 'Testing 5' }
  const content6 = { type: 'post', text: 'Testing 6' }
  const content7 = { type: 'post', text: 'Testing 7' }
  let state = validate.initial()
  state = validate.appendNew(state, null, keys, content0, Date.now())
  state = validate.appendNew(state, null, keys, content1, Date.now() + 1)
  state = validate.appendNew(state, null, keys, content2, Date.now() + 2)
  state = validate.appendNew(state, null, keys, content3, Date.now() + 3)
  state = validate.appendNew(state, null, keys, content4, Date.now() + 4)
  state = validate.appendNew(state, null, keys, content5, Date.now() + 5)
  state = validate.appendNew(state, null, keys, content6, Date.now() + 6)
  state = validate.appendNew(state, null, keys, content7, Date.now() + 7)

  const offset0 = (await addMsgPromise(state.queue[0].value, log)).offset
  const offset1 = (await addMsgPromise(state.queue[1].value, log)).offset
  const offset2 = (await addMsgPromise(state.queue[2].value, log)).offset
  const offset3 = (await addMsgPromise(state.queue[3].value, log)).offset
  const offset4 = (await addMsgPromise(state.queue[4].value, log)).offset
  const offset5 = (await addMsgPromise(state.queue[5].value, log)).offset
  const offset6 = (await addMsgPromise(state.queue[6].value, log)).offset
  const offset7 = (await addMsgPromise(state.queue[7].value, log)).offset

  // IMPORTANT! Causes the bitvector index to have offset1 and offset2
  await pify(jitdb.prepare)(slowEqual('value.content.type', 'post'))

  await pify(log.del)(offset1)
  t.pass('delete msg 1')
  await pify(log.del)(offset2)
  t.pass('delete msg 2')
  await pify(log.onDeletesFlushed)()

  const source = query(
    fromDB(jitdb),
[...]

PR #225, opened 2022-07-18
fix pagination nextSeq when there are holes

staltz:
Fixes (somewhat) issue #223.

`jitdb.paginate`'s "nextSeq" calculation is improved now, by looking at all the seqs, not at the non-deleted records (this would have cause the same "page" to be scanned a few times, resulting in duplicates).

And now `toPullStream()` can emit pages that are smaller than the requested page size, but it will never emit pages of size zero, we pull.filter those out.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1192,12 +1192,12 @@ module.exports = function (log, indexesPath) {
     seq = seq || 0
 
     const sorted = sortedBy(bitset, descending, sortBy)
-    const resultSize = sorted.size
+    const total = sorted.size
 
     // seq -> record buffer
     const recBufferCache = {}
 
-    function processResults(seqs, resultSize) {
+    function processResults(seqs) {
       push(
         push.values(seqs),
         push.asyncMap((seq, continueCB) => {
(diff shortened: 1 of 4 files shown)


arj03:
Converting the code into how I see it, to see if I got the changes right:

So this makes paginate behave like limit, in that it can return fewer results if things are deleted?

There is still the compaction problem, but I do think this is a good step forward in that we have one less problem to worry about.

staltz:
> So this makes paginate behave like limit, in that it can return fewer results if things are deleted?

Yes



> There is still the compaction problem, but I do think this is a good step forward in that we have one less problem to worry about.

What do you mean with "the compaction problem" (I've been solving so many problems, I get lost which one we're talking about)

arj03:
> What do you mean with "the compaction problem" (I've been solving so many problems, I get lost which one we're talking about)?

:grin: this one: https://github.com/ssbc/jitdb/pull/224#discussion_r921961951

staltz:
Ah, right. I call that the rewinding problem. :sweat_smile: 

Yes, I'll revise that part.

Issue #226, opened 2022-07-18
seq rewinding in paginate() is incorrect


staltz:
Rewinding `seq - holesFound` is wrong, there's actually many ways how the rewinding is wrong.

E.g. 1: If `seq` is 3 but there were 1 deletes at seq 10, then you don't need to rewind, but we still do.

E.g. 2: if `seq` is 3 but there were 100 deletes at seq 200–300, then, we're going to rewind to seq -97, which is nonsensical.

But more fundamentally, the problem is that a `paginate()` that is run automatically post-compaction is using a `seq` argument referring to the previous uncompacted version of the log, and this `seq` argument has no commitment to the new realities of the compacted log. **The seq system of the log before and after compaction is completely rewritten**, because the log file itself is rewritten, so old values `offset` and `seq` are irrelevant.

staltz:
@arj03 This is a hairy problem, and I'm thinking we can just bail out instead of solving it. We can fix `toPullStream` so that it listens to compaction, and as soon as compaction is done, we can **end** the pull stream (I'm not sure if it should end with error or `true`, but anyway, ends).

staltz:
Okay, I found another solution, and it goes like this:

- After every `paginate()` is done, keep track of the msg.key for the latest msg in the page
- Compaction starts and the next `paginate()` call gets put on the queue, waiting for compaction to end
- When compaction ends, look up the `msg.key` for the latest msg, **find the seq** for that msg.key, and then resume `paginate()` with that seq+1 (we can safely assume that seq+1 is non-deleted because *nothing* is deleted after compaction)

On paper, it should work, but in practice there's quite a hassle to get `msg.key` => seq because that's something recorded in ssb-db2, not jitdb.

arj03:
I think it would be ok to just end the stream and let the other end continue if they want. Not sure how big of a hassle it is?

staltz:
It is a hassle for `live({ old: true })` streams because in that case we would end the *old* part of the stream, but the *live* part would continue as is. Consumers of the stream would receive a few old items, and then suddenly jump to receive only live items, and they would miss some important old items.

PR #227, opened 2022-07-18
keep state file to signal reindex post-compaction

staltz:
This PR solves 3 problems:

- Issue #226
- `reindex` was wrong for bitset indexes
- Keeping track of when should `reindex()` be called post-compact should be a jitdb concern
  - (Currently ssb-db2 persists three state files for this, one of them for jitdb reindexing)

These changes should cause a **breaking change** (because of the added argument in `paginate()`) and we should also remove the ssb-db2 `post-compact-reindex-jit` state file.

### DIFF
--- a/index.js
+++ b/index.js
@@ -121,7 +121,7 @@ module.exports = function (log, indexesPath) {
       function conclude() {
         EmptyFile.delete(postCompactReindexPath, () => {
           if (waitingCompaction.length === 0) return
-          for (const cb of waitingCompaction) cb(stats.holesFound)
+          for (const cb of waitingCompaction) cb()
           waitingCompaction.length = 0
         })
       }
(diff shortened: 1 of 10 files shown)


index.js
```diff
@@ -1519,6 +1592,8 @@ module.exports = function (log, indexesPath) {
           }
         }
 
+        if (index.bitset) index.bitset.removeRange(seq, Infinity)
```
staltz, line 1595: This was a whoops

arj03: WOAH

index.js
```diff
@@ -1251,17 +1285,59 @@ module.exports = function (log, indexesPath) {
...
+    )
+  }
+
   function countBitsetSlice(bitset, seq, descending) {
     if (!seq) return bitset.size()
     else return bitset.size() - seq
   }
 
-  function paginate(operation, seq, limit, descending, onlyOffset, sortBy, cb) {
+  function paginate(
```
arj03, line 1317: The README needs update for this :)

index.js
```diff
@@ -1251,17 +1285,59 @@ module.exports = function (log, indexesPath) {
...
+    operation,
+    seq,
+    limit,
+    descending,
+    onlyOffset,
+    sortBy,
+    latestMsgKey,
```
arj03, line 1324: Can we call this compactionLatestMsgKey or something like that? To signal that this is only ever used for compaction.

arj03:
end of review. Overall this approach looks good.

staltz:
Thanks for taking a look! I updated it now.

arj03 (approved):
:star2:

PR #229, opened 2022-07-18
use Map for indexes dict

staltz:
Somewhat similar to #228, this refactor uses Map for the `indexes` dictionary. I haven't measured, but it could be that we get some tiny perf benefits too.

(This is the last PR before I release jitdb v7)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 /.vscode
 pnpm-lock.yaml
 *~
+.nyc_output
+coverage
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -51,33 +54,37 @@ module.exports = function (log, indexesPath) {
...
         count: 0,
         tarr: new Uint32Array(16 * 1000),
         version: 1,
-      }
+      })
     }
 
+    seqIndex = indexes.get('seq')
+    timestampIndex = indexes.get('timestamp')
+    sequenceIndex = indexes.get('sequence')
```
staltz, line 86: Using `seqIndex` directly instead of `indexes['seq']` or `indexes.get('seq')` could be a performance improvement. Not a big one, but given that we have `indexes['seq']` on hot code path, this could make a difference.

PR #232, opened 2022-09-01
optimize reindex() when seq is zero

staltz:
Just a neat optimization I stumbled upon.


test/reindex.js
```diff
@@ -187,7 +187,14 @@ prepareAndRunTest('reindex prefix map', dir, (t, db, raf) => {
...
           t.equal(results.length, 2)
-          t.end()
+
+          const secondMsgOffset = 352
+          db.reindex(secondMsgOffset, () => {
+            db.all(typeQuery, 0, false, false, 'declared', (err, results) => {
+              t.equal(results.length, 2)
+              t.end()
+            })
+          })
```
staltz, line 197: This is just to guarantee that we cover all the cases inside reindex()

arj03 (approved):
Looks good :)

PR #236, opened 2022-12-05
Make live queries aware of reindexed (decrypted) values

arj03:
With this https://github.com/ssbc/ssb-db2/pull/410 is now working. Clearly this needs a test here as well. But first I was wondering if this would be a good approach or not?

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ const pull = require('pull-stream')
 const mutexify = require('mutexify')
 const toPull = require('push-stream-to-pull-stream')
 const pullAsync = require('pull-async')
+const pullMany = require('pull-many')
 const TypedFastBitSet = require('typedfastbitset')
 const bsb = require('binary-search-bounds')
 const multicb = require('multicb')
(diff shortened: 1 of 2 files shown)


staltz:
Thanks for tackling this!

What do you think about a new API for this? I'm wondering if adding these decrypted msgs into the live() will break some assumptions about our APIs. Two assumptions: (1) every message appears in the stream only once, (2) live stream delivers only msgs recently added to the log.

arj03:
That is a good point. I considered that, but as you can see this is only in live. So for this to happen you would have to receive a message and the key at almost the same time. I fear that a new apo might be a gotcha.

staltz:
Wouldn't it be possible to read the same msg from the live stream, even if receival time is very different to decryption time?

Like if I subscribe to this live stream at 13:00, then receive a box2 encrypted msg at 13:02, and then later I'm added to a group at 14:00 and thus I get the decrypted msg at 14:00.

What I'm worried is that we're do a breaking change on live() and then creating new issues somewhere downstream. It's easier to make a non-breaking change with a new API.

staltz:
> maybe a good compromise would be to have two streams as you say and then people can combine them if needed.

I like that compromise.

> It's an interesting question about semantics slightly_smiling_face like what did the original live promise

From our readme, it says (bold is my emphasis):

> Will setup a pull stream and this in cb. The pull stream will emit new values **as they are added to the underlying log**. This is meant to run after paginate or all.

arj03:
Yeah, I'm working on that solution. It's almost ready, just have to get all the tests working. It's a pitty that live has this old: true flag, in that when we add another operator (like decrypted()) it gets a bit muddy what these mean when you combine them. But the diff is pretty short. Will share later when I get everything running properly.

staltz:
Also, if we split the stream then shouldn't this be implemented in ssb-db2? We can pick only the records that have been box2 decrypted and pass them into some pull-stream.

I actually don't understand how this PR #236 works, because it's adding the lowest offset of the range being reindexed, but we should be streaming all the offsets that have been *decrypted* within that range.

staltz:
Sorry for yet another comment, but

> I actually don't understand how this PR https://github.com/ssbc/jitdb/pull/236 works, because it's adding the lowest offset of the range being reindexed, but we should be streaming all the offsets that have been decrypted within that range.

I see now how it works because https://github.com/ssbc/ssb-db2/blob/b095c33cde3be59ce6e6e69be47ce80fdaf8116e/core.js#L1015 is called for all the just-decrypted records, BUT shouldn't we change this implementation such that we only call jitdb.reindex for the **minimum** offset, such that this offset covers the whole range of decrypted records? See e.g. that Airtable task about optimizing reindexEncrypted

arj03:
Thanks for the input, working on a better solution now.

PR #238, opened 2023-01-13
replace push.collect with push.drain

staltz:
**Problem:** `push.collect` accumulates an array of all results, but we don't actually need or use those results, while this could be a source of memory consumption that could contribute to out-of-memory crashes.

**Solution:** use `push.drain` instead.


arj03 (approved):
👍

PR #241, opened 2023-01-28
ignore DEFERRED ops in prepare()

staltz:
## Context

In Manyverse I use `jitdb.prepare()` to warm up all the jit indexes, and I recently added `prepare` on `isPrivate('box')`:

https://github.com/staltz/manyverse/blob/781f6066e4fdfa3451b8c9ce48b4235d0374b8e3/src/backend/plugins/dbUtils.ts#L195

Which in ssb-db2 may use a `deferred`:

https://github.com/ssbc/ssb-db2/blob/289e00a964be367c882accb0a131576136c5b602/operators/index.js#L181-L197

## Problem

In jitdb, we only resolve/execute the `DEFERRED` tasks in `operators.js` and if the developer uses either `toCallback()` or `toPullStream()`. But when using `prepare()`, we actually don't use `toCallback()` nor `toPullStream()`. This means that the `DEFERRED` operation ends up in `executeOperation` and bumps into this line:

https://github.com/ssbc/jitdb/blob/3bb1da79e52f60567f3d42f92ab4772023ad7171/index.js#L994

This doesn't cause a crash or any stressful problem, but it causes a `console.error` that may be annoying.

## Solution

If we have a `DEFERRED` in our `prepare()` then just execute the deferred operation.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,6 +44,7 @@ module.exports = function (log, indexesPath) {
   let timestampIndex
   let sequenceIndex
   let isReady = false
+  let jitdb
   const waitingReady = new Set()
   let compacting = false
   let compactStartOffset = null
@@ -991,7 +992,25 @@ module.exports = function (log, indexesPath) {
     } else if (!op.type) {
       // to support `query(fromDB(jitdb), toCallback(cb))`
       getFullBitset(cb)
-    } else console.error('Unknown type', op)
+    } else if (op.type === 'DEFERRED') {
+      // DEFERRED only appears in this pipeline when using `prepare()` API
+      op.task(
+        { jitdb },
+        (err, newOp) => {
+          if (err) {
+            console.error('Error executing DEFERRED operation', err)
+            cb(null, new TypedFastBitSet())
+            return
+          }
+          if (newOp) {
+            getBitsetForOperation(newOp, cb)
+          } else {
+            cb(new TypedFastBitSet())
+          }
+        },
+        function onAbort() {}
+      )
+    } else console.error('Unknown type in jitdb executeOperation:', op)
   }
 
   function forEachLeafOperationIn(operation, fn) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -991,7 +992,25 @@ module.exports = function (log, indexesPath) {
...
       // to support `query(fromDB(jitdb), toCallback(cb))`
       getFullBitset(cb)
-    } else console.error('Unknown type', op)
+    } else if (op.type === 'DEFERRED') {
+      // DEFERRED only appears in this pipeline when using `prepare()` API
+      op.task(
+        { jitdb },
+        (err, newOp) => {
+          if (err) {
+            console.error('Error executing DEFERRED operation', err)
```
arj03, line 1001: why not bubble up the error here?

index.js
```diff
@@ -991,7 +992,25 @@ module.exports = function (log, indexesPath) {
...
+            cb(null, new TypedFastBitSet())
+            return
+          }
+          if (newOp) {
+            getBitsetForOperation(newOp, cb)
+          } else {
+            cb(new TypedFastBitSet())
```
arj03, line 1008: should be: `null, new TypedFastBitSet()`, right?

index.js
staltz: The `cb` is not `cb(err, result)`, it's `cb(bitset, filters)`, unfortunately

index.js
staltz: Same as above :)
cb is `cb(bitset, filters)`

index.js
arj03: right, then this is fine. But would probably be good to change :)

arj03 (approved):
Reviewed in person :+1:

staltz:
I'm going to also revert that `let jitdb` that isn't needed anymore, and rename the PR as well.

